PR #3, opened 2023-05-19
chore: Add build dependencies to packages

ling1726:
All packages are built with @swc/helpers and tslib. Add these dependencies to every package.json in build output

### DIFF
--- a/packages/nx-plugin/src/executors/build/executor.ts
+++ b/packages/nx-plugin/src/executors/build/executor.ts
@@ -60,11 +60,8 @@ function copyPackageJson(paths: PackagePaths) {
       ...packageJson.dependencies,
     },
     peerDependencies: {
-      '@fluentui/react-components': '>=9.0.0 <10.0.0',
       '@types/react': '>=16.8.0 <19.0.0',
-      '@types/react-dom': '>=16.8.0 <19.0.0',
       react: '>=16.8.0 <19.0.0',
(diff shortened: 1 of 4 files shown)


packages/nx-plugin/src/executors/build/executor.ts
```diff
@@ -54,6 +55,19 @@ function copyPackageJson(paths: PackagePaths) {
       type: 'git',
       url: 'https://github.com/microsoft/fluentui-contrib',
     },
+    dependencies: {
+      tslib: '^2.3.0',
+      '@swc/helpers': '~0.5.1',
```
layershifter, line 60: Why both? It should be only one...

ling1726: you're right, this is cruft...

packages/nx-plugin/src/executors/build/executor.ts
```diff
@@ -54,6 +55,19 @@ function copyPackageJson(paths: PackagePaths) {
...
+      ...packageJson.dependencies,
+    },
+    peerDependencies: {
+      '@fluentui/react-components': '>=9.0.0 <10.0.0',
```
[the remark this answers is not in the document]
ling1726: you're right, I'll move that back and start from 9.20.0 which is the current version in this repo

layershifter (approved):

PR #9, opened 2023-06-16
feat: introduce @fluentui-contrib/react-shadow

layershifter:
This package introduces `@fluentui-contrib/react-shadow`.

Work in progress, see for more details https://github.com/microsoft/fluentui/issues/27759


packages/react-shadow/src/renderers/createShadowDOMRenderer.ts
```diff
@@ -0,0 +1,204 @@
...
+  sheetToInsert: T,
+  targetSheet: T | null
+): T[] {
+  if (targetSheet === null) {
+    return [...arr, sheetToInsert];
+  }
+
+  const index = arr.indexOf(targetSheet);
+
+  return [...arr.slice(0, index + 1), sheetToInsert, ...arr.slice(index + 1)];
```
ling1726, line 112:
```suggestion
  return arr.splice(index, 0, sheetToInsert);
```

should do the same ting

ling1726 (approved):

packages/react-shadow/src/renderers/createShadowDOMRenderer.ts
```diff
@@ -0,0 +1,204 @@
...
+          const sheet = getCSSStyleSheetForBucket(
+            cssSheetsCache,
+
+            styleBucketName,
+            metadata,
+
+            (styleSheet) => {
+              const targetStyleSheet = findInsertionPoint(renderer, styleSheet);
+
+              renderer.adoptedStyleSheets = insertBefore(
```
ling1726, line 150: if griffel stylesheets could somehow keep references to each other, there's no need to maintain a copy of the relative order of adopted style sheets

packages/react-shadow/src/renderers/createShadowDOMRenderer.ts
spmonahan: Since the return value is set on `.adoptedStyleSheets` this function must return a new array reference.

https://developer.mozilla.org/en-US/docs/Web/API/Document/adoptedStyleSheets#append_a_new_stylesheet

layershifter: This is a requirement for Chrome below 99, after that version `.adoptedStyleSheets` are mutable.

PR #15, opened 2023-06-21
Set tabindex=0 and use document role for chat with focusable content

adamsamec:
#### Description of changes

This PR fullfuls the intent of the original PR which was meant to make chat message content focusable by setting tabindex="0". It also replaces the not yet existing aria-readingmode attribute with the role="document" attribute and adds role="application" attribute for the &lt;Chat&gt; component.


tsconfig.base.json
```diff
@@ -16,6 +16,9 @@
     "strict": true,
     "baseUrl": ".",
     "paths": {
+      "@fluentui-contrib/accessible-chat": [
+        "packages/accessible-chat/src/index.ts"
+      ],
```
ling1726, line 21:
```suggestion
```

ling1726 (approved):

PR #20, opened 2023-07-04
fix(react-shadow): render portals under shadow root

layershifter:
Adopts the change from https://github.com/microsoft/fluentui/pull/28395 and renders components powered by Portals in shadow roots.

### DIFF
--- a/packages/react-shadow/stories/root/Portals.stories.tsx
+++ b/packages/react-shadow/stories/root/Portals.stories.tsx
@@ -1,0 +1,80 @@
+import { root } from '@fluentui-contrib/react-shadow';
+import {
+  Button,
+  makeStyles,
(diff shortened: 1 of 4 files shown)


packages/react-shadow/stories/root/Portals.stories.tsx
```diff
@@ -0,0 +1,80 @@
...
+};
+
+export const Portals = () => (
+  <div
+    style={{
+      border: '3px dotted orange',
+      padding: 10,
+      display: 'flex',
+      flexDirection: 'column',
+      gap: 10,
```
ling1726, line 65: nit: also use griffel for this, all stories are decorated with a FluentProvider

ling1726 (approved):

PR #23, opened 2023-07-04
chore: Add extraneous dependency eslint rule

ling1726:
Adds the
[no-extraneous-dependency](https://github.com/import-js/eslint-plugin-import/blob/HEAD/docs/rules/no-extraneous-dependencies.md) eslint rule so that there is a guarantee that all dependencies are declared in package.json.

Fixes the violations in react-chat and nx-plugin

### DIFF
--- a/packages/nx-plugin/package.json
+++ b/packages/nx-plugin/package.json
@@ -4,5 +4,11 @@
   "version": "0.0.1",
   "type": "commonjs",
   "generators": "./generators.json",
-  "executors": "./executors.json"
+  "executors": "./executors.json",
+  "dependencies": {
+    "semver": "^7.5.2",
+    "@nx/devkit": "^16.1.4",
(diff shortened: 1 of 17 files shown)


packages/nx-plugin/package.json
```diff
@@ -4,5 +4,11 @@
...
   "type": "commonjs",
   "generators": "./generators.json",
-  "executors": "./executors.json"
+  "executors": "./executors.json",
+  "dependencies": {
+    "semver": "^7.5.2",
+    "@nx/devkit": "^16.1.4",
+    "@nx/storybook": "^16.1.4",
+    "@nx/js": "16.1.4"
+  }
```
layershifter, line 13: Is there sense to lint private packages for deps?

ling1726: I can't see why it would hurt. The intention in the future is to publish this so that others can also build their own repo

PR #25, opened 2023-07-05
chore: Add verify-deps executor to verify peer deps

ling1726:
Adds a new executor `verify` deps that will copy individual packages to a tmp directory to build them with typescript with the minimum peer dependency versions. If the build fails, it should be an indicator that peer dependencies are not correct.

**This check will not correctly handle cases where dependencies are within the monorepo. Create an issue to track on completion of this PR**


packages/nx-plugin/src/executors/verify-deps/executor.ts
```diff
@@ -0,0 +1,110 @@
...
+      pkgJson.dependencies[depName] = minVersion.version;
+    }
+  }
+
+  // Copy all types dependencies - they are not that big and there's not many of them (yet)
+  const rootPkgJson = readJsonFile(path.join(workspaceRoot, 'package.json'));
+  const rootDevDeps = Object.entries(rootPkgJson.devDependencies ?? {});
+  pkgJson.devDependencies ??= {};
+  for (const [depName, version] of rootDevDeps) {
+    if (depName.startsWith('@types/')) {
```
layershifter, line 104: This will also copy these, correct?

https://github.com/microsoft/fluentui-contrib/blob/823224320f87b291d794c0e18da4de9c4b5cf5a0/package.json#L45-L47

packages/nx-plugin/src/generators/library/generator.ts
```diff
@@ -58,10 +62,7 @@ export default async function (tree: Tree, options: LibraryGeneratorSchema) {
 
     packageJson.peerDependencies ??= {
       '@fluentui/react-components': `>=${reactComponentsVersion} <10.0.0`,
-      '@types/react': '>=16.8.0 <19.0.0',
-      '@types/react-dom': '>=16.8.0 <19.0.0',
```
layershifter, line 62: Is this intentional?

packages/nx-plugin/src/executors/verify-deps/executor.ts
ling1726: yeah, there's no real way to find out what types pacakges are necessary without doing an AST parse of the src

layershifter (approved):

packages/nx-plugin/src/executors/verify-deps/executor.ts
layershifter: I would rather do just `@types/react` & `@types/react-dom` instead. This should cover all use cases.

packages/nx-plugin/src/executors/verify-deps/executor.ts
```diff
@@ -0,0 +1,110 @@
...
+  if (!projectsConfigurations || !projectName) {
+    throw new Error('no project configurations');
+  }
+
+  const { root } = projectsConfigurations.projects[projectName];
+  const { name: tmpDir, removeCallback: cleanup } = tmp.dirSync({
+    prefix: projectName,
+    unsafeCleanup: true,
+  });
+  console.info('copying', projectName, 'to', tmpDir);
```
layershifter, line 25: I also suggest to do an early return if a package does not have `peerDependencies`.

layershifter (approved):

ling1726:
Closing because the method of this verification is incorrect - the verification needs to be done by _**consuming**_ the package

PR #32, opened 2023-07-07
React data grid react window/fix/align peer dependencies

cymushier:
Align peer dependencies to enable strict package managers know and link correct dependencies.


ling1726 (commented):
These dependencies are already applied to the package json during build following #31

ling1726 (changes requested):
This or is not necessary

layershifter:
@cymushier FYI, this is completely wrong behavior, https://github.com/VincentBailly/midgard-yarn-strict/issues/9

PR #37, opened 2023-08-14
feat(react-data-grid-react-window): allow overriding virtualized list props

sjwilczynski:
In scope of this PR we add a new prop to the component `listProps` which will be passed through to `FixedListSize` component that is leveraged for virtualization. It allows partner teams to pass some custom properties to virtualized list, e.g some stying. We add an example through Storybook showcasing how style overrides could be used to achieve better UX for users.

Note to reviewer: I was surprised that adding package as devDependency in `react-data-grid-react-window` didn't result in lock file change and I ended adding it in root package.json. Do we want to maybe use [workspaces](https://classic.yarnpkg.com/lang/en/docs/workspaces/)?


layershifter (approved):
I would rather expose a slot to be consistent with Fluent, but this suggestion includes non-trivial typings problems that makes me think that it's not a good idea. We also _probably_ don't want to allow much flexibility there.

I am fine with `listProps`, but I will let @ling1726 make the final review.

layershifter:
> Note to reviewer: I was surprised that adding package as devDependency in `react-data-grid-react-window` didn't result in lock file change and I ended adding it in root package.json. Do we want to maybe use [workspaces](https://classic.yarnpkg.com/lang/en/docs/workspaces/)?

This repository uses Nx, so it does not need to have workspaces configured. @ling1726 it's probably a good idea to add an explainer about dependency management in this repository to `CONTRIBUTING.md`.

sjwilczynski:
Thanks for feedback. Then I will wait for @ling1726 feedback and in case something should be done with dependencies (as I understand keeping them in root is a feature :)) let me know. Looking forward to getting this merged!

packages/react-data-grid-react-window/stories/DataGrid/VirtualizedDataGrid.stories.tsx
```diff
@@ -241,3 +243,82 @@ VirtualizedDataGrid.parameters = {
...
+  },
+  list: {
+    scrollbarGutter: 'stable',
+  },
+  grid: {
+    height: '100%',
+  },
+});
+
+export const ReactWindowOverrides = () => {
```
ling1726, line 260: There is no automation to enforce this yet, but move this example to a separate `.stories.tsx` file, otherwise the code sample on the published storybook can be quite funny

change/@fluentui-contrib-react-data-grid-react-window-79465f5b-778b-4d2a-ae97-3fe844a4ec79.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "add capability to override FixelListSize props",
```
ling1726, line 3:
```suggestion
  "comment": "feat: add capability to override FixelListSize props",
```

packages/react-data-grid-react-window/stories/DataGrid/VirtualizedDataGrid.stories.tsx
sjwilczynski: Done :)

sjwilczynski:
Adjusted based on comments, @ling1726 let me know if we are ready to merge

ling1726 (approved):

PR #39, opened 2023-09-01
feat: add HeadlessFluentProvider to @fluentui-contrib/react-headless-provider

layershifter:
This PR adds `HeadlessFluentProvider` under `@fluentui-contrib/react-headless-provider`.

This is the same as `FluentProvider` from `@fluentui/react-components`, however it does not render additional any DOM elements, only applies React Contexts needed for Fluent UI components to render.

Following features work:
- rendering into different documents
- `customStyleHooks_unstable` & `overrides_unstable` props

Following features don't work:
- scoped theming i.e. theming should be configured globally and separately
- scoped "dir" i.e. text direction should be configured globally
- polyfill for `:focus-visible` (`useFocusVisible()` from `@fluentui/react-tabster`) should be initialized separately

Designed to be used with `createCSSRuleFromTheme()` (https://github.com/microsoft/fluentui/pull/29052), see for context https://github.com/microsoft/fluentui/issues/21290.

### Usage

```tsx
import { HeadlessFluentProvider } from "@fluentui-contrib/react-headless-provider"

function App() {
  return (
    <HeadlessFluentProvider targetDocument={someDocument}>
      {/* children */}
    </HeadlessFluentProvider>
  ) 
}

```


packages/react-headless-provider/README.md
```diff
@@ -0,0 +1,11 @@
+# react-headless-provider
```
ling1726, line 1: Please update the README

ling1726 (approved):

PR #43, opened 2023-10-23
feat: Implement stylelint plugin

ling1726:
Implements a stylelint plugin for Fluent UI specific rules. Starts off
with 2 rules:

* combinator-depth
* no-focus-visible


packages/stylelint-plugin/src/create-rule.ts
```diff
@@ -0,0 +1,23 @@
+import { Rule, RuleMessages, RuleBase } from 'stylelint';
```
layershifter, line 1:
```suggestion
import type { Rule, RuleMessages, RuleBase } from 'stylelint';
```

packages/stylelint-plugin/src/postcss-utils.ts
```diff
@@ -0,0 +1,31 @@
+import type { Document, AtRule, Root, Container, Rule } from 'postcss';
+export function isAtRule(container: { type: string }): container is AtRule {
```
layershifter, line 2:
```suggestion
import type { Document, AtRule, Root, Container, Rule } from 'postcss';

export function isAtRule(container: { type: string }): container is AtRule {
```

packages/stylelint-plugin/src/rules/combinator-depth/README.md
```diff
@@ -0,0 +1,51 @@
...
+id or a CSS class directly on the affected DOM element.
+
+## Incorrect
+
+```tsx
+import { makeStyles, Button, buttonClassNames } from '@fluentui/react-components';
+
+const useStyles = makeStyles({
+  root: {
+    [`.${buttonClassNames.root}`]: {
```
layershifter, line 14:
```suggestion
    [`& .${buttonClassNames.root}`]: {
```

packages/stylelint-plugin/src/rules/combinator-depth/combinator-depth.ts
```diff
@@ -0,0 +1,68 @@
...
+      postcssResult,
+      ruleName,
+      {
+        actual: allowedDepth,
+        possible: [(val) => typeof val === 'number'],
+      }
+    );
+
+    if (!validOptions) {
+      return;
```
layershifter, line 30: Should not it throw?

packages/stylelint-plugin/src/rules/no-focus-visible/no-focus-visible.test.ts
```diff
@@ -0,0 +1,69 @@
+import rule from './no-focus-visible';
+import * as stylelint from 'stylelint';
+
+describe('no-focus-visible', () => {
+  it('should fail with :focus-visible', async () => {
+    const { errored, results } = await stylelint.lint({
+      code: `.foo:focus-visible { color: red; }`,
```
layershifter, line 7: Let's add a fixture with `.foo :focus-visible` to ensure that space is handled, too.

packages/stylelint-plugin/src/rules/combinator-depth/combinator-depth.ts
```diff
@@ -0,0 +1,68 @@
...
+
+        continue;
+      }
+      tokenizedSelector.reverse();
+      let combinatorCount = 0;
+      for (const token of tokenizedSelector) {
+        if (token.type === 'combinator') {
+          combinatorCount++;
+        }
+      }
```
layershifter, line 55:
```suggestion
      const combinatorCount = tokenizedSelector.filter(token.type === 'combinator').length;
```

?

packages/stylelint-plugin/src/rules/no-focus-visible/no-focus-visible.ts
```diff
@@ -0,0 +1,72 @@
...
+        )
+      ) {
+        stylelint.utils.report({
+          ruleName,
+          message: messages.focusVisible(selector),
+          messageArgs: [selector],
+          result: postcssResult,
+          node: selectorToRule[selector],
+        });
+      } else if (
```
layershifter, line 59:
```suggestion
        });
        continue;
      }
      
      if (
```

packages/stylelint-plugin/src/rules/no-focus-visible/no-focus-visible.ts
```diff
@@ -0,0 +1,72 @@
...
+    if (!validOptions) {
+      return;
+    }
+
+    const { selectorToRule, selectors } = extractAllSelectors(postcssRoot);
+
+    for (const selector of selectors) {
+      let tokenizedSelector = [];
+      try {
+        tokenizedSelector = parsel.tokenize(selector);
```
layershifter, line 34: Do you really need to tokenize there? Can't we simply use `.indexOf(":focus-visible")`?

layershifter (approved):

packages/stylelint-plugin/src/rules/no-focus-visible/no-focus-visible.ts
ling1726: personally I prefer `else if` over `continue`

PR #45, opened 2023-10-25
chore: fix type export maps

ling1726:
Fixes the type export maps path


ling1726:
closing as duplicate of #46

PR #60, opened 2024-01-03
Implement "insertion point" API for `react-shadow`

### LINKED ISSUE #59: Add support for "insertion point" to `react-shadow`
Griffel has an [`insertionPoint`](https://griffel.js.org/react/api/create-dom-renderer#insertionpoint) API that allows users to control where Griffel styles are inserted into a document. This is useful for controlling style specificity via `<style>` tag order. For example, users can ensure all their application styles come after Griffel styles, thus ensuring app styles "win" any specificity battle based on document order (all else being equal of course).

`react-shadow` should implement a similar API as the current implementation always places all Griffel styles after all application styles meaning application styles always "lose" specificity battles. This comes up in real-world Web Component applications that embed `react-shadow` as these applications have their own styles and would like to apply class names to Fluent controls without having to re-implement styles as `makeStyles()` calls. [A basic example](https://stackblitz.com/edit/stackblitz-starters-xe4vlu?file=src%2FApp.tsx).

spmonahan:
Adds an "insertion point" API similar to [Griffel's API](https://griffel.js.org/react/api/create-dom-renderer#insertionpoint).

Here we use a special sentinel `ExtendedCSSStyleSheet` in the `adoptedStyleSheets` array and look it up prior to setting the stylesheets.

~Additionally, a constant for whether `adoptedStyleSheets` is mutable is added so we can conditionally update the array. This improves Safari support. [See this PR](https://github.com/microsoft/fast/pull/6703) for more details.~

Resolves #59


packages/react-shadow/src/root.tsx
```diff
@@ -5,7 +5,15 @@ import {
...
 import { createProxy } from 'react-shadow';
+import {
+  REACT_SHADOW_INSERTION_BUCKET_NAME,
+  SUPPORTS_MODIFYING_ADOPTED_STYLESHEETS,
+} from './constants';
 import type { Root } from 'react-shadow';
+import type { GriffelShadowDOMRenderer } from '@griffel/shadow-dom';
+
+// Hack to get this type that is not directly exported at the moment.
+type ExtendedCSSStyleSheet = GriffelShadowDOMRenderer['adoptedStyleSheets'][0];
```
ling1726, line 16: Is it possible to create this type once and reuse it? I see it's also being used in `makeInsertionPointSheet`

packages/react-shadow/src/util.ts
```diff
@@ -0,0 +1,12 @@
+import { REACT_SHADOW_INSERTION_BUCKET_NAME } from './constants';
+import type { GriffelShadowDOMRenderer } from '@griffel/shadow-dom';
+
+// Hack to get this type that is not directly exported at the moment.
+type ExtendedCSSStyleSheet = GriffelShadowDOMRenderer['adoptedStyleSheets'][0];
+
+export const makeInsertionPointSheet = (): ExtendedCSSStyleSheet => {
```
ling1726, line 7: since app CSSStylesheets need to be created, why is it necessary to make the user create a new sentinel stylesheet? couldn't they simply pass the last stylesheet they append as the insertion point?

You could then find the insertion point by a simple equality check without any extra metadata. It's also easier in userland since they don't need to call any specific utility to make this work

packages/react-shadow/src/root.tsx
```diff
@@ -24,10 +32,33 @@ const ReactComponentsWrapper: React.FC<{
...
-      ];
+      const insertionIndex = root.adoptedStyleSheets.findIndex(
+        (styleSheet) =>
+          (styleSheet as ExtendedCSSStyleSheet).metadata?.[
+            REACT_SHADOW_INSERTION_BUCKET_NAME
+          ] === true
+      );
+
+      if (insertionIndex !== -1) {
+        if (SUPPORTS_MODIFYING_ADOPTED_STYLESHEETS) {
```
ling1726, line 43: mutable stylesheets introduce and interesting problem - if users prepend/append stylesheets this effect will never run because the reference to the adopted stylesheet never changes.

However currently there is a bunch of conditions at the start of the effect that essentially stop any updates to adopted stylesheets anyway...

spmonahan: We should probably work out how this bit works because it also addresses a bug in Safari: https://github.com/microsoft/fast/pull/6703

That said, maybe it's best to drop it from this PR and follow up later?

packages/react-shadow/src/root.tsx
spmonahan: Yes. It is actually defined in `@griffel/shadow-dom` and, imo, should be exported from there but I didn't want to get this PR bogged down in getting packages aligned :)

packages/react-shadow/src/util.ts
spmonahan: I like this idea (it's where I originally started), but I couldn't see a good way to pass the sentinal into `root`. Will consider it more.

packages/react-shadow/src/root.tsx
```diff
@@ -24,10 +32,33 @@ const ReactComponentsWrapper: React.FC<{
         return;
       }
 
-      root.adoptedStyleSheets = [
-        ...root.adoptedStyleSheets,
-        ...renderer.adoptedStyleSheets,
-      ];
```
layershifter, line 30: TL;DR This change should be ported to Griffel, the relevant place is `findInsertionPoint()`: https://github.com/microsoft/griffel/blob/f134bc6056599a78cf27b71dec854224a162ee22/packages/shadow-dom/src/findInsertionPoint.ts#L9

----

As I recall, this effect is solving the issue of syncing sheets from `react-shadow` and Griffel as the logic in `react-shadow` will simply erase them:

https://github.com/Wildhoney/ReactShadow/blob/1432b818cf0ba96b8a86920d4b426eb41911358c/src/core/index.js#L60-L61

However, now that effect will re-create also `root`, so new Griffel renderer will be created. I created a PR to remove this logic in the first place (#61).

packages/react-shadow/src/root.tsx
spmonahan: Since some other work needs to happen in Griffel I'm updating there to export the type.

packages/react-shadow/src/root.tsx
```diff
@@ -6,12 +6,19 @@ import { createShadowDOMRenderer } from '@griffel/shadow-dom';
 import * as React from 'react';
 import { createProxy } from 'react-shadow';
 import type { Root } from 'react-shadow';
+import type { ExtendedCSSStyleSheet } from '@griffel/shadow-dom';
```
spmonahan, line 9: Note: this import depends on https://github.com/microsoft/griffel/pull/480

packages/react-shadow/src/root.tsx
spmonahan: I've removed this for later consideration.

packages/react-shadow/src/util.ts
spmonahan: With the current `createProxy()` API I don't think there is a way to do this unfortunately. Perhaps @layershifter has an idea though?

spmonahan:
> As we discussed in offline, let's introduce a factory to create `root`, i.e.:
> 
> ```tsx
> const root = createRoot({ sheetInsertionPoint: sheet })
> ```

@layershifter Made this update.

yarn.lock
```diff
@@ -2249,7 +2249,7 @@
   dependencies:
     "@swc/helpers" "^0.5.1"
 
-"@griffel/core@^1.12.2", "@griffel/core@^1.14.1", "@griffel/core@^1.14.3":
+"@griffel/core@^1.14.1", "@griffel/core@^1.14.3":
```
layershifter, line 2252: Dedupe these?

layershifter (approved):

yarn.lock
spmonahan: Looks like those dupes a pre-existing and stem from `@fluentui/react-provider` and `@griffel/react`, neither of which come from this PR. Maybe dedupe in a separare PR?

layershifter: image

@spmonahan you won't be able to pass CI without deduping...

spmonahan: I guess that's a good reason to fix it now ;)

spmonahan:
@layershifter I think I've addressed the changes we discussed. Can you remove your hold on the PR so it can be merged?

layershifter:
> @layershifter I think I've addressed the changes we discussed. Can you remove your hold on the PR so it can be merged?

@spmonahan I have already approved the PR previous week 🐱 

image

@microsoft/fluentui-react-build can you please check the PR? 🙏

packages/nx-plugin/src/generators/library/generator.spec.ts
```diff
@@ -33,7 +33,7 @@ describe('create-package generator', () => {
     const pkgJson = readJson(tree, paths.packageJson);
     expect(pkgJson.peerDependencies).toMatchInlineSnapshot(`
       {
-        "@fluentui/react-components": ">=9.46.3 <10.0.0",
+        "@fluentui/react-components": ">=9.46.5 <10.0.0",
```
Hotell, line 36: this should be probably a separate PR ? I dont see need for v-build to approve lot of implementation details that we dont have context about.

Hotell (approved):

PR #65, opened 2024-01-12
React resize handle

george-cz:
Say goodbye to you static layouts, they are so 2023!

It's time to welcome react-resize-handle package to fill your layout resizing needs. Just add the refs to your HTML elements and observe your layout changes in real time, without any react re-renders at all!

Docs/tests coming later after the approach has been validated internally.

This is still very much in prototype/POC/before internal review phase and should not be used by anyone, ever.


packages/react-resize-handle/src/hooks/useMouseHandler.ts
```diff
@@ -0,0 +1,144 @@
...
+  const globalWin = targetDocument?.defaultView;
+
+  const startCoords = React.useRef({ clientX: 0, clientY: 0 });
+  const { growDirection, onValueChange, elementRef } = params;
+
+  const key =
+    growDirection === 'right' || growDirection === 'left' ? 'width' : 'height';
+  const initialValue = React.useRef(
+    elementRef.current?.getBoundingClientRect()[key] || 0
+  );
```
layershifter, line 34:
```suggestion
  const initialValue = React.useRef(elementDimension(elementRef.current, growDirection));
```

??

packages/react-resize-handle/src/hooks/useMouseHandler.ts
```diff
@@ -0,0 +1,144 @@
...
+    }
+  );
+
+  const onDrag = useEventCallback((event: NativeTouchOrMouseEvent) => {
+    //Using requestAnimationFrame improves performance on slower CPUs
+    if (typeof globalWin?.requestAnimationFrame === 'function') {
+      requestAnimationFrame(() => recalculatePosition(event));
+    } else {
+      recalculatePosition(event);
+    }
```
layershifter, line 82:
```suggestion
    targetWindow?.requestAnimationFrame(() => recalculatePosition(event));
```

I curious why this condition is needed at all: https://developer.mozilla.org/en-US/docs/Web/API/window/requestAnimationFrame

packages/react-resize-handle/src/hooks/useMouseHandler.ts
```diff
@@ -0,0 +1,144 @@
...
+      targetDocument?.removeEventListener('touchend', onDragEnd);
+      targetDocument?.removeEventListener('touchmove', onDrag);
+    }
+
+    params.onDragEnd?.(event);
+  });
+
+  const onPointerDown = useEventCallback((event: NativeTouchOrMouseEvent) => {
+    startCoords.current = getEventClientCoords(event);
+    initialValue.current = elementDimension(elementRef.current, growDirection);
```
layershifter, line 101: Naming of `startCoords` and `initialValue` is confusing, can we name it better?

packages/react-resize-handle/src/hooks/useMouseHandler.ts
```diff
@@ -0,0 +1,144 @@
...
+          break;
+        case 'left':
+          newValue -= deltaCoords[0] * (dir === 'rtl' ? -1 : 1);
+          break;
+        case 'up':
+          newValue -= deltaCoords[1];
+          break;
+        case 'down':
+          newValue += deltaCoords[1];
+          break;
```
layershifter, line 58:
```suggestion
      switch (growDirection) {
        case 'left':
        case 'right':
          newValue -= deltaCoords[0] * (dir === 'rtl' ? -1 : 1);
          break;
        case 'up':
          newValue -= deltaCoords[1];
          newValue += deltaCoords[1];
          break;
      }
```

?

packages/react-resize-handle/src/hooks/useMouseHandler.ts
```diff
@@ -0,0 +1,144 @@
...
+
+      onValueChange(Math.round(newValue));
+
+      // If, after resize, the element size is different than the value we set, that we have reached the boundary
+      // and the element size is controlled by something else (minmax, clamp, max, min css functions etc.)
+      // In this case, we need to update the value to the actual element size so that the css var and a11y props
+      // reflect the reality.
+      const elSize = Math.round(
+        elementRef.current?.getBoundingClientRect()[key] || 0
+      );
```
layershifter, line 69:
```suggestion
  const elSize = elementDimension(elementRef.current, growDirection);

```

??

layershifter:
@george-cz LGTM, left some minor comments

packages/react-resize-handle/stories/Default.stories.tsx
```diff
@@ -0,0 +1,141 @@
...
+  const pageStyles = usePageStyles();
+  const wrapperStyles = useMainWrapperStyles();
+  const boxStyles = useMainBoxStyles();
+
+  const [maxValue, setMaxValue] = React.useState(400);
+
+  const {
+    handleRef: navHandleRef,
+    wrapperRef: navWrapperRef,
+    elementRef: navElementRef,
```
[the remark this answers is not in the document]
george-cz: I'm actually expecting that these will be called at the wrapper level and passed in a context (based on my testing in TMP), and then handleRef and ElementRef will be consumed somewhere else. Adding the refs key doesn't really add anything, imho

packages/react-resize-handle/src/hooks/useMouseHandler.ts
george-cz: this doesnt look right?

packages/react-resize-handle/package.json
```diff
@@ -4,6 +4,8 @@
   "private": true,
   "peerDependencies": {
     "@fluentui/react-components": ">=9.35.1 <10.0.0",
+    "@fluentui/react-utilities": ">=9.15.1 <10.0.0",
+    "@fluentui/react-shared-contexts": ">=9.7.2 <10.0.0",
```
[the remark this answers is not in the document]
layershifter: Okay, then please declare is as a normal dependency.

packages/react-resize-handle/src/hooks/useMouseHandler.ts
layershifter: Yeah, oops

packages/react-resize-handle/package.json
```diff
@@ -4,6 +4,7 @@
   "private": true,
```
layershifter, line 4: You probably also want to make it public to be released.

layershifter (approved):

PR #73, opened 2024-01-17
feat(tooling): update CODEOWNERS when creating a new library/package

### LINKED ISSUE #68: feat(nx-plugin): setup CODEOWNER within `library` generator
ATM its not straightforward to understand what's the approval process for newly created packages.

we have setup codeowners which is answer to former question.  This is alread in core fluent, so we need to propagate that functionality here as well.

marcosmoura:
The process of updating CODEOWNERS was manual. With this PR, it is now mandatory to add one when creating a new package/library.
This takes as inspiration code from fluentui main repository.

Please note that this code should be removed once we have the unified tooling in place.

Fixes #68


packages/nx-plugin/src/generators/library/schema.json
```diff
@@ -12,7 +12,26 @@
...
+      "$default": {
+        "$source": "argv",
+        "index": 1
+      },
+      "x-priority": "important",
+      "x-prompt": {
+        "message": "Which team will own this library",
+        "type": "list",
+        "items": [
+          "@microsoft/cxe-red",
```
Hotell, line 28: because this is contrib/extensions repo owners might vary. would be great if we could add "another" option so user can specify a custom team/owner if necessary. 

as a start I'd remove the x-prompt as type of list to force user always manually add owning team.

marcosmoura: That makes sense!

Hotell (approved):

PR #78, opened 2024-01-18
feat(react-draggable-dialog): create draggable dialog with handle

marcosmoura:
Creates the base logic for a functional Draggable Dialog.
Tests will be added later to split the logic into two PRs.

The code is large as the components are too coupled tied (because of a shared context), so in case a review gets confusing, I can gladly explain all the bits.


packages/react-draggable-dialog/package.json
```diff
@@ -2,11 +2,16 @@
   "name": "@fluentui-contrib/react-draggable-dialog",
   "version": "0.0.1",
   "private": true,
+  "dependencies": {
+    "@dnd-kit/core": "^6.1.0",
+    "@dnd-kit/utilities": "^3.2.2",
+    "@dnd-kit/modifiers": "^7.0.0"
```
layershifter, line 8: image


FYI: I haven't seen that bits of dnd-kit are used, but in some scenario it has extremely poor performance, https://github.com/clauderic/dnd-kit/issues/1194#issuecomment-1696704815

packages/react-draggable-dialog/src/contexts/DraggableDialogContext.ts
```diff
@@ -0,0 +1,31 @@
...
+  id: string;
+  isDragging: boolean;
+  hasBeenDragged: boolean;
+  position: {
+    x: number;
+    y: number;
+  };
+};
+
+export const draggableDialogContextDefaultValue = {
```
Hotell, line 13: this shouldn't be exported right ?

```suggestion
const draggableDialogContextDefaultValue = {
```

packages/react-draggable-dialog/package.json
```diff
@@ -2,11 +2,16 @@
   "name": "@fluentui-contrib/react-draggable-dialog",
   "version": "0.0.1",
   "private": true,
+  "dependencies": {
+    "@dnd-kit/core": "^6.1.0",
```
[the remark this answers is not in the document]
marcosmoura: They are all the latest. `@dnd-kit/core` is the main library and the other two are "extensions" that are compatible with the main, so it can be different.

packages/react-draggable-dialog/src/components/DraggableDialog/DraggableDialog.test.tsx
```diff
@@ -4,6 +4,10 @@ import { DraggableDialog } from './DraggableDialog';
 
 describe('DraggableDialog', () => {
   it('should render', () => {
-    render(<DraggableDialog />);
+    render(
```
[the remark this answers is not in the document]
marcosmoura: This is not supposed to be a full test, but rather a blank file ready to receive the test code. I have them done, but I'm splitting into a separate PR to make this one smaller.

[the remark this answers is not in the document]
Hotell: IMO when adding logic the test should be part of the PR, otherwise we are merging something that we cannot guarantee to work heh.

but I'll leave this up to your judgement , resolving

packages/react-draggable-dialog/src/components/DraggableDialog/DraggableDialog.types.ts
```diff
@@ -0,0 +1,68 @@
...
+    /**
+     * A initial y coordinate of the draggable dialog.
+     */
+    y?: number;
+  };
+
+  /**
+   * Text to be announced by screen readers when the dialog is dragged.
+   */
+  announcements?: {
```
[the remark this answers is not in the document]
marcosmoura: The partners might not need announcements as an internal requirement, but they should be available and be part of an accessible Drag'n drop solution
These announcements here are necessary, to translate the text that the underlying library outputs.

package.json
```diff
@@ -55,22 +58,22 @@
     "@typescript-eslint/eslint-plugin": "^5.58.0",
     "@typescript-eslint/parser": "^5.58.0",
     "beachball": "^2.33.2",
-    "eslint": "~8.15.0",
     "eslint-config-prettier": "8.1.0",
     "eslint-plugin-import": "^2.27.5",
-    "jest": "^29.4.1",
+    "eslint": "~8.15.0",
```
Hotell, line 63: I assume this re-ordering happened by `yarn add` ?

packages/react-draggable-dialog/src/hooks/useDraggableDialog.ts
```diff
@@ -0,0 +1,5 @@
+import { useDraggableDialogContext } from '../contexts/DraggableDialogContext';
```
Hotell, line 1: thought: any reason why this i not part of contexts ? this seems unnecessary separation by wrong type of entity, while from domain POV this should live with context.

packages/react-draggable-dialog/src/components/DraggableDialogHandle/DraggableDialogHandle.types.ts
```diff
@@ -0,0 +1,10 @@
+import * as React from 'react';
+
+export type DraggableDialogHandleProps = React.HTMLAttributes<HTMLElement> & {
+  /**
+   * The element that will act as the handle for dragging the dialog.
+   * Only one child is allowed.
+   * @required
+   */
+  children: JSX.Element;
```
Hotell, line 9: using `JSX` global namespace is not the most robust approach for specifying types. 

can we use `React.ReactElement` ?

packages/react-draggable-dialog/src/components/DraggableDialog/utils.ts
```diff
@@ -0,0 +1,77 @@
...
+        bottom: windowRect.bottom + margin.bottom,
+        left: windowRect.left + margin.start,
+      },
+    });
+  };
+
+export const getParsedDraggableMargin = (
+  margin: DraggableMargin | undefined
+): Required<DraggableMarginViewport> => {
+  const defaultMargin = {
```
Hotell, line 41: this doesn't need to be re-created on every invocation. can we move defaults constant to the module scope ?

packages/react-draggable-dialog/src/components/DraggableDialog/useDraggableDialog.ts
```diff
@@ -0,0 +1,177 @@
...
+    };
+  }
+
+  return {
+    ...defaultMargin,
+    ...margin,
+  };
+};
+
+export const useDraggableDialog = (_opts: DraggableDialogOptions) => {
```
Hotell, line 72: what's the deal with `_` prefix ?

packages/react-draggable-dialog/src/components/DraggableDialog/useDraggableDialog.ts
```diff
@@ -0,0 +1,177 @@
...
+      let { x, y } = {
+        x: event.clientX - currentPosition.current.x,
+        y: event.clientY - currentPosition.current.y,
+      };
+
+      if (!keepInViewport) {
+        return { x, y };
+      }
+
+      if (x - margin.start + offsetLeft < 0) {
```
Hotell, line 98: im wondering if all those checks are necessary to evaluate on every function call ?

packages/react-draggable-dialog/src/components/DraggableDialog/DraggableDialog.tsx
```diff
@@ -1,8 +1,91 @@
...
+    setIsDragging(true);
+  }, []);
+
+  const restrictToMargin = React.useMemo(() => {
+    return restrictToMarginModifier({ margin, keepInViewport });
+  }, [margin, keepInViewport]);
+
+  const dndAnnouncements = React.useMemo(() => {
+    if (!announcements) {
+      return;
```
Hotell, line 60: can we change this api to be more explicit ? return `null` ?

packages/react-draggable-dialog/src/components/DraggableDialog/DraggableDialog.tsx
```diff
@@ -1,8 +1,91 @@
...
+    return {
+      onDragStart: () => announcements.start,
+      onDragEnd: () => announcements.end,
+    };
+  }, [announcements]);
+
+  return (
+    <DndContext
+      sensors={sensors}
+      modifiers={[restrictToMargin]}
```
Hotell, line 72: haven't check the latest development in performance for jsx rendering, but AFAIR these inline objects will be re-created on every render, and because we are instantiating context all children using the context included. can we improve this ?

packages/react-draggable-dialog/src/components/DraggableDialog/DraggableDialog.types.ts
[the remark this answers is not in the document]
Hotell: right. I dont see any default values for these, in order to enforce a11y experiences should these be required ?

packages/react-draggable-dialog/src/components/DraggableDialog/DraggableDialog.test.tsx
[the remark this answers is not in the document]
Hotell: unresolving ->

quite a lot logic was added to this PR , It would be great to add these tests as well as part of it. I dont feel comfortable to approve such amount of logic without any automatic verification that it works.

packages/react-draggable-dialog/src/components/DraggableDialogSurface/DraggableDialogSurface.types.ts
```diff
@@ -0,0 +1,10 @@
+import { DialogSurfaceProps } from '@fluentui/react-components';
+
+export type DraggableDialogSurfaceProps = DialogSurfaceProps & {
+  /**
+   * The element that will be draggable.
+   * Only one child is allowed.
+   * @required
+   */
+  children: JSX.Element;
```
Hotell, line 9:
```suggestion
  children: React.ReactElement;
```

packages/react-draggable-dialog/src/components/DraggableDialog/useDraggableDialog.ts
marcosmoura: This file was included by mistake, as it used to contain logic that was separated and placed in the right places. I removed it.

packages/react-draggable-dialog/src/components/DraggableDialog/DraggableDialog.tsx
```diff
@@ -1,8 +1,91 @@
...
+    return restrictToMarginModifier({ margin, keepInViewport });
+  }, [margin, keepInViewport]);
+
+  const dndAnnouncements = React.useMemo(() => {
+    if (!announcements) {
+      return;
+    }
+
+    return {
+      onDragStart: () => announcements.start,
```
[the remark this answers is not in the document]
marcosmoura: Actually the `DndContext` has sane defaults for when `announcements` is not provided. The reason for those `announcements` prop is to have a way to translate the default message. By default, it announces in english.

packages/react-draggable-dialog/src/components/DraggableDialog/DraggableDialog.tsx
marcosmoura: Fixed!

packages/react-draggable-dialog/src/components/DraggableDialog/DraggableDialog.tsx
marcosmoura: We can't set as null, because that'll mean that we don't want any accessibility props. `undefined` will count as "we didn't pass any value".

package.json
marcosmoura: Yes. I reverted as this is unnecessary.

packages/react-draggable-dialog/src/hooks/useDraggableDialog.ts
marcosmoura: More logic was meant to be added, but for simplification, I reverted this change. It now lives with the context.

packages/react-draggable-dialog/src/components/DraggableDialog/useDraggableDialog.ts
marcosmoura: This file is now removed.

packages/react-draggable-dialog/src/components/DraggableDialogSurface/DraggableDialogSurface.types.ts
marcosmoura: This is actually unnecessary, as `DialogSurfaceProps` already has `children`. I removed this explicit type.

packages/react-draggable-dialog/src/components/DraggableDialogHandle/DraggableDialogHandle.types.ts
marcosmoura: Fixed.

packages/react-draggable-dialog/src/components/DraggableDialog/DraggableDialog.types.ts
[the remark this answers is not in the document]
marcosmoura: I don't know if we need it, to be honest. The DndKit library has [defaults](https://github.com/clauderic/dnd-kit/blob/master/packages/core/src/components/Accessibility/defaults.ts) for the narration, which will happen regardless of the usage of this prop. The value of this prop here is to provide a way to translate those default messages, which are in English. Partners that don't need to fulfil the narration requirement would have to provide with their defaults for those values, which they don't need.
What do you think about this?

packages/react-draggable-dialog/src/components/DraggableDialog/DraggableDialog.tsx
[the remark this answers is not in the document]
Hotell: I see . well that's the hit n miss with 3rd party libraries as some/most things are not obvious in a PR. thank you!

packages/react-draggable-dialog/src/components/DraggableDialog/DraggableDialog.types.ts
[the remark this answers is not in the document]
Hotell: yeah we can keep it as is. I'd consider though to provide those defaults within our codebase to be explicit.

Hotell (approved):

PR #80, opened 2024-01-19
chore(react-tree-grid): initial implementation

bsunderhus:
1. removes @adamsamec example
2. adds initial minimal implementation to be discussed


packages/react-tree-grid/package.json
```diff
@@ -5,6 +5,7 @@
   "peerDependencies": {
     "@fluentui/react-components": ">=9.44.4 < 10.0.0",
     "@fluentui/react-shared-contexts": ">=9.7.2 <10.0.0",
+    "@fluentui/react-context-selector": ">=9.1.47 < 10.0.0",
     "@fluentui/keyboard-keys": ">=9.0.6 < 10.0.0",
     "@fluentui/react-tabster": ">=9.14.0 < 10.0.0",
     "@fluentui/react-utilities": ">=9.15.1 < 10.0.0",
```
layershifter, line 11: Please declare all of these as dependencies instead: there is no expectation that user will install these packages on their side, but will be forced (or noticed) by a package manager.

packages/react-tree-grid/package.json
```diff
@@ -5,6 +5,7 @@
   "peerDependencies": {
     "@fluentui/react-components": ">=9.44.4 < 10.0.0",
     "@fluentui/react-shared-contexts": ">=9.7.2 <10.0.0",
```
layershifter, line 7: This package package should not be defined in `peerDependencies` -  all bits should be coming from `@fluentui/react-components`. Anyway, I don't see any usage of this package anymore.

packages/react-tree-grid/src/components/TreeGrid/TreeGrid.types.ts
```diff
@@ -0,0 +1 @@
+export type TreeGridProps = JSX.IntrinsicElements['div'];
```
layershifter, line 1: image

packages/react-tree-grid/src/components/TreeGrid/TreeGrid.tsx
```diff
@@ -1,20 +1,18 @@
...
-const useStyles = makeResetStyles({
-  display: 'block',
-});
+import { mergeClasses } from '@fluentui/react-components';
+import { useTreeGridStyles } from './useTreeGridStyles.styles';
+import { TreeGridProps } from './TreeGrid.types';
+import { useNavigation } from '../../hooks/useNavigation';
 
 export const TreeGrid = React.forwardRef(
   (props: TreeGridProps, ref: React.ForwardedRef<HTMLDivElement>) => {
```
layershifter, line 8: This component does not follow the structure of Fluent components. What is the plan there?

packages/react-tree-grid/src/components/TreeGridRow/TreeGridRow.tsx
```diff
@@ -1,25 +1,34 @@
...
 import {
-  mergeClasses,
-  useTableRowStyles_unstable,
-  useTableRow_unstable,
-  TableRowState,
-} from '@fluentui/react-components';
-
-export type TreeGridRowProps = JSX.IntrinsicElements['div'];
+  useArrowNavigationGroup,
+  useFocusableGroup,
```
layershifter, line 7: These should come from `@fluentui/react-components`

packages/react-tree-grid/src/hooks/useNavigation.ts
```diff
@@ -0,0 +1,97 @@
...
+  return { onKeyDown: handleKeyDown, ...tabsterAttributes };
+};
+
+const useFindParentRow = () => {
+  const { findPrevFocusable } = useFocusFinders();
+  return React.useCallback(
+    (currentRow: HTMLElement): HTMLElement | null => {
+      const currentLevel = Number(currentRow.getAttribute('aria-level'));
+      if (isNaN(currentLevel)) {
+        console.error(
```
layershifter, line 73: Let's wrap with `env !== production`

layershifter (approved):

packages/react-tree-grid/src/components/TreeGrid/TreeGrid.tsx
bsunderhus: None of the components in fluent contrib does

packages/react-tree-grid/src/components/TreeGridRow/TreeGridRow.types.ts
```diff
@@ -0,0 +1,4 @@
+export type TreeGridRowProps = JSX.IntrinsicElements['div'] & {
+  // aria-level is required for screen readers to understand the nesting level of the row
+  'aria-level': number;
```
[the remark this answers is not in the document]
bsunderhus: None, we could let it be `number | string`, I'm just forcing it to not be optional

PR #84, opened 2024-01-23
Scaffolds "pierce-dom" package

spmonahan:
`pierce-dom` is a package for "shadow DOM aware" implementations of DOM APIs. For example, a `TreeWalker` that walks into shadow roots, a `MutationObserver` that observes inside shadow roots, a `getActiveElement()` function that gets the `activeElement` inside shadow roots and so on.

This will be used by Tabster but the actual implementation should be separate as it is likely to be useful outside of Tabster itself.

For examples of how this will be used see
- https://github.com/microsoft/tabster/pull/347
- https://github.com/microsoft/fluentui/pull/30286


Hotell:
related - it looks like tabster should became a monorepo as I see no valid reason to keeping closely related things in separate repositories:

- keyborg
- tabster
- pierce-dom
- other a11y related packages

Hotell:
pipeline is failing because false positives provided in storybook generator - workin on fix

spmonahan:
> it looks like tabster should became a monorepo as I see no valid reason to keeping closely related things in separate repositories:
> 
> * keyborg
> * tabster
> * pierce-dom
> * other a11y related packages
> 
> cc @microsoft/teams-prg ?

I'm fine with the idea of Tabster being a monorepo but are keyborg and pierce-dom really part of Tabster? Seems like they are useful for Tabster but also useful independent of it.

Hotell:
> > it looks like tabster should became a monorepo as I see no valid reason to keeping closely related things in separate repositories:
> > 
> > * keyborg
> > * tabster
> > * pierce-dom
> > * other a11y related packages
> > 
> > cc @microsoft/teams-prg ?
> 
> I'm fine with the idea of Tabster being a monorepo but are keyborg and pierce-dom really part of Tabster? Seems like they are useful for Tabster but also useful independent of it.

FWIW keyborg is a big part of tabster 

image

based on your description of this package IMO it fits nicely into the scope of "tools for accessible navigation on web" ;) 

ideally even react-tabster should exist there or any particular `<your-js-framework>-tabster`

does it make sense ?

Hotell (approved):
approving, once scaffold is updated an pipeline will pass

spmonahan:
> based on your description of this package IMO it fits nicely into the scope of "tools for accessible navigation on web" ;)
> 
> ideally even react-tabster should exist there or any particular `<your-js-framework>-tabster`
> 
> does it make sense ?

Yeah, I get where you're coming from. My point was that keyborg (to my understanding) and pierce-dom are designed specifically to be independent of Tabster itself and bucketing them together with Tabster might make it seem that they are designed only for use with Tabster when they offer utility without it.

Hotell:
> That said, I'm fine with a monorepo for "accessible web navigation" but for the moment we have immediate need of this package so let's do the monorepo update later.

yeah, there is no rush on this for sure. just wanted to get folks POV on potential future plans.

PR #85, opened 2024-01-23
feat(react-tree-grid): implements TreeGridRowGroup component

bsunderhus:
## Changes

1. Introduces `TreeGridRowGroup` component

## Possible problems with this implementation

1. there's nothing limiting the user to add more than one `groupOwner` in the same group
2. there's nothing limiting that the first row on a `TreeGridRowGroup` should be the `groupOwner`
3. each `TreeGridRow` controls it's own visibility (this might not be ideal for virtualization cases, as the visibility of the nodes is dictated by the virtualization library on scroll normally?!)

### DIFF
--- a/packages/react-tree-grid/package.json
+++ b/packages/react-tree-grid/package.json
@@ -3,13 +3,13 @@
   "version": "0.0.1",
   "private": true,
   "peerDependencies": {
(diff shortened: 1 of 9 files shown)


bsunderhus:
Blocked by https://github.com/microsoft/fluentui-contrib/pull/86

layershifter (approved):

PR #88, opened 2024-01-23
chore(nx-plugin): add floating promises lint rules in order to catch nasty bugs

Hotell:
- migrate all generator tempaltes to  have `.template` extension which is stripped away by nx devkit
  - this also makes tooling setup more easier as otherwise these templates are consumed by eslint/tsc etc which is wrong
- introduces crucial Promise related lint rules for `nx-plugin` project
  - "@typescript-eslint/await-thenable": "error",
  - "@typescript-eslint/no-misused-promises": "error",
  - "@typescript-eslint/no-floating-promises": "error" 
- fixes exposed critical runtime issue in one of ours executors


packages/nx-plugin/src/executors/type-check/executor.ts
```diff
@@ -34,7 +34,7 @@ export default async function runExecutor(
   });
 
   if (errors?.length) {
-    printDiagnostics(errors);
```
Hotell, line 37: you shall not pass next time ! ;)

micahgodbolt (approved):

ling1726 (approved):

PR #90, opened 2024-01-30
chore: nx migrate to v17.3

bsunderhus:


.nx/cache/10410386269260275797.commit
```diff
@@ -0,0 +1 @@
+true
```
spmonahan, line 1: Should `.nx/` be in `.gitignore`?

packages/pierce-dom/.babelrc
```diff
@@ -0,0 +1,3 @@
+{
+  "extends": "../../babel.config.json"
```
ling1726, line 2: shouldn't the storybook generator be configured to add this to every new package?

change/@fluentui-contrib-react-data-grid-react-window-grid-25562c5b-7f46-4097-8b83-8b50cfbdf43a.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "patch",
```
Hotell, line 2: all of these changes should be type none. this should trigger any version bumps

packages/nx-plugin/src/generators/configure-storybook/files/.babelrc.template
```diff
@@ -0,0 +1,3 @@
+{
+  "extends": "../../babel.config.json"
```
Hotell, line 2: this shouldn't be here as that has nothing to do with storybook, this nicely shows that we need library generator for react and node

packages/nx-plugin/src/generators/configure-storybook/generator.ts
```diff
@@ -16,9 +16,8 @@ export default async function (
 
   const { root: projectRoot } = project;
   await configurationGenerator(tree, {
-    name,
+    project: name,
     configureCypress: false,
-    uiFramework: '@storybook/react-webpack5',
```
Hotell, line 21: wondering why is this remove ? based on API it's required https://nx.dev/nx-api/storybook/generators/configuration#uiframework

packages/nx-plugin/src/generators/configure-storybook/files/.babelrc.template
bsunderhus: I added this here, as it's failing on `storybook-build` command... what do you suggest?

Hotell: right, yeah I know why you added it there. can you create issue to introduce a `react` library generator ? that's where some of the "misused" logic of storybook config generator belongs

bsunderhus: https://github.com/microsoft/fluentui-contrib/issues/93

Hotell (approved):

.gitignore
```diff
@@ -37,3 +37,6 @@ testem.log
 # System Files
 .DS_Store
 Thumbs.db
+
+# nx
+.nx
```
spmonahan, line 42: I know I suggested this but I think it should actually be

```suggestion
.nx/cache
```

based on https://github.com/microsoft/fluentui/blob/master/.gitignore#L161

@Hotell should be able to confirm

packages/nx-plugin/src/generators/configure-storybook/files/.babelrc.template
Hotell: yes , thanks !

.gitignore
Hotell: nx migration did this, I trust robots ;) :D lets keep it as is

ling1726 (approved):

marcosmoura (approved):

spmonahan (approved):

PR #91, opened 2024-01-30
useShadowDOMSupport() hook.

mshoho:


spmonahan (approved):

packages/pierce-dom/package.json
```diff
@@ -2,6 +2,9 @@
   "name": "@fluentui-contrib/pierce-dom",
   "version": "0.0.1",
   "private": true,
+  "dependencies": {
+    "tabster": "^6.0.0"
```
ling1726, line 6: NX is not actually a monorepo in the sense that each package manages its own dependencies. 

You'll need to install tabster 6.x in dev dependencies at the root of this repo.

`yarn add -D tabster@6`

ling1726: Please follow up with an update in Fluent UI so that we can remove the explicit tabster install

PR #96, opened 2024-02-02
feat(react-tree-grid): adds virtualization support

bsunderhus:
1. deletes `TreeGridRowGroup` component in favor of `subtree` slot
    * The component is problematic for the virtualization scenario (as for virtualization libraries the rendering should be flattened, with all components in the same level and with `TreeGridRowGroup` this is not possible
    * Some problems with this implementation were pointed here https://github.com/microsoft/fluentui-contrib/pull/85
2. adds virtualization story (similar to FlatTree)


Hotell (approved):
approving for build related files

packages/react-tree-grid/package.json
```diff
@@ -11,8 +11,12 @@
   },
   "dependencies": {
     "@fluentui/react-context-selector": ">=9.1.47 < 10.0.0",
+    "@fluentui/react-jsx-runtime": ">=9.0.29 < 10.0.0",
     "@fluentui/keyboard-keys": ">=9.0.6 < 10.0.0",
     "@fluentui/react-tabster": ">=9.14.0 < 10.0.0",
     "@fluentui/react-utilities": ">=9.16.0 < 10.0.0"
+  },
+  "devDependencies": {
+    "react-window": "^1.8.10"
```
Hotell, line 20: why is this needed ? package.json shouldn't contain any `devDependendies` definitions as that has nothing to do with production release.

packages/react-tree-grid/package.json
```diff
@@ -11,6 +11,7 @@
   },
   "dependencies": {
     "@fluentui/react-context-selector": ">=9.1.47 < 10.0.0",
+    "@fluentui/react-jsx-runtime": ">=9.0.29 < 10.0.0",
```
Hotell, line 14: wondering why are we using these kind of semver ranges when it can be simplified to `^9.0.29` ?

Hotell (approved):
approving for v-build related files

packages/react-tree-grid/package.json
bsunderhus: I have the faintest idea. I'm just following the pattern I've seen in the other packages, @ling1726 might know more?!

ling1726: They are equivalent. Personally I just got into the habit of using that syntax to declare peer dep. Since generally, peer deps go across several majors.

Hotell: yeah for peerDeps its valid, but for unlikely for deps

packages/react-tree-grid/stories/Default.stories.tsx
```diff
@@ -17,111 +16,117 @@ import {
...
-          </TreeGridCell>
-          <TreeGridCell>
-            <Menu>
-              <MenuTrigger disableButtonEnhancement>
-                <Button>Toggle menu</Button>
-              </MenuTrigger>
+      <TreeGridRow
+        subtree={
+          <>
+            <TreeGridRow>
```
ling1726, line 22: Not sure if you're going to do a documentation pass later too, but would be nice to make these subtrees separate components

packages/react-tree-grid/stories/Virtualization.stories.tsx
```diff
@@ -0,0 +1,243 @@
...
+      <TreeGridCell header>{item.children}</TreeGridCell>
+      <TreeGridCell aria-colspan={3}>
+        <Button>Header action</Button>
+      </TreeGridCell>
+    </TreeGridRow>
+  ) : (
+    <TreeGridRowProvider
+      value={{
+        open: !!openItems.get(item.parentValue),
+        level: 1,
```
ling1726, line 107:
```suggestion
        level: 2,
```

should this be the next level?

ling1726 (approved):
The slotted subtree isn't too appealing for me, but I can't think of better and all the points you raised were valid

packages/react-tree-grid/stories/Default.stories.tsx
ling1726: maybe have an example too with manual `aria-level` set? It seems that for the case we are designing this for, it should be feasible to abstract theses to domain specific examples

```ts
const Header = <TreeGridRow aria-level="1" />

const Content = <TreeGridRow aria-level="2" />

<Header />
<Content />
<Content />
<Content />
<Content />
<Header />
<Content />
<Content />
```

bsunderhus: We can definitely have an example without using `subtree` and letting the user manually control depth, but unfortunately this is more than just setting `aria-level`. It's also about controlling visibility, so this example that you just gave would be a little bit more complex than just that 😅

PR #100, opened 2024-02-12
chore(react-tree-grid): adds playwright component tests

bsunderhus:
1. adds playwright component test configuration
2. adds playwright test for TreeGrid keyboard/click behaviours

### DIFF
--- a/packages/react-tree-grid/jest.config.ts
+++ b/packages/react-tree-grid/jest.config.ts
@@ -1,5 +1,6 @@
 /* eslint-disable */
 import { readFileSync } from 'fs';
+import { Config } from 'jest';
 
 // Reading the SWC compilation config and remove the "exclude"
 // for the test files to be compiled by SWC
@@ -18,7 +19,7 @@ if (swcJestConfig.swcrc === undefined) {
 // jest needs EsModule Interop to find the default exported setup/teardown functions
 // swcJestConfig.module.noInterop = false;
 
-export default {
+const config: Config = {
   displayName: 'button',
   preset: '../../jest.preset.js',
   transform: {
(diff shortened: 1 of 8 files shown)


Hotell:
once this lands to main https://github.com/microsoft/fluentui-contrib/pull/102

- re-generate pw config

Hotell:
we need to discuss file name patterns, because with current setup if project would use jest and pw component test, the only distinction ATM is `.test` vs `.spec` but jest actually invokes also `.spec` which would explode.

with cypress we use `*.cy.*` suffix, maybe we can do similar for playwright, something like `.pw.tsx` etc

bsunderhus:
> once this lands to main #102
> 
> * re-generate pw config

@Hotell,

Trying to follow up on this, I'm getting some errors while trying to test locally

image

I went to `main` branch and regenerated the playwright configuration and  a spec test, and this happens whenever I try to run `component-test`

Hotell:
ups, while in rush I created invalid options generation dictionary 

fix https://github.com/microsoft/fluentui-contrib/pull/116

packages/react-tree-grid/jest.config.ts
```diff
@@ -18,12 +19,15 @@ if (swcJestConfig.swcrc === undefined) {
...
-export default {
+const config: Config = {
   displayName: 'button',
   preset: '../../jest.preset.js',
   transform: {
     '^.+\\.[tj]sx?$': ['@swc/jest', swcJestConfig],
   },
   moduleFileExtensions: ['js', 'ts', 'tsx', 'html'],
   coverageDirectory: '../../coverage/packages/button',
+  modulePathIgnorePatterns: ['.spec.tsx'],
```
Hotell, line 30: this will hit us sooner than later. I'll create issue to track this

Hotell: https://github.com/microsoft/fluentui-contrib/issues/118

Hotell (approved):

PR #95, opened 2024-02-01
feature(nx-plugin): adds playwright executor and generators

bsunderhus:
Adds `playwright` for component test according to [documentation](https://playwright.dev/docs/test-components)

1. executor:
    * `playwright` (if `testingType === component` then execute task manually, otherwise delegate to `@nx/playwright`)
2. generators:
    * `playwright-component-configuration` (as per https://nx.dev/nx-api/react/generators/cypress-component-configuration)
    * `component-playwright-spec` (as per https://nx.dev/nx-api/react/generators/component-cypress-spec)

PR #102, opened 2024-02-12
feat(nx-plugin): redo playwright executor/generator setup to accept all nx flags and make it work

Hotell:
**Before:**
our custom playwright executor wasn't working before.

**After:**

After detailed analysis of core nx/playwright plugin I realized that there are various additional gymnastics that need to happen which goes beyond simple `execSync` call.

Most of the implementation was copied from nx/playwright with our custom API modification + pointing to proper CLI for component test.

Last but not least CI needs to install playwright browsers so our executor will now handle this behind the scenes.

Generator was tweaked to incorporate new target changes including eslint setup to follow playwright best practices

## Related issues:

Follows https://github.com/microsoft/fluentui-contrib/pull/95

### DIFF
--- a/packages/nx-plugin/src/generators/playwright-component-configuration/files/playwright.config.ts.template
+++ b/packages/nx-plugin/src/generators/playwright-component-configuration/files/playwright.config.ts.template
@@ -1,6 +1,12 @@
 import { defineConfig, devices } from '@playwright/experimental-ct-react';
 
 /**
+ * Read environment variables from file.
+ * https://github.com/motdotla/dotenv
+ */
+// require('dotenv').config();
+
+/**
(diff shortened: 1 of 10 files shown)


packages/nx-plugin/src/generators/playwright-component-configuration/files/playwright.config.ts.template
```diff
@@ -1,5 +1,11 @@
 import { defineConfig, devices } from '@playwright/experimental-ct-react';
 
+/**
```
Hotell, line 3: nx generator adds this as well, for consistency I added it here as well

bsunderhus (approved):

tudorpopams (approved):

PR #107, opened 2024-02-19
test(react-draggable-dialog): add basic test cases

marcosmoura:
- Includes a non-production warning when rendering `DraggableDialogHandle` and `DraggableDialogSurface` outside a `DraggableDialog` as they don't work without it.
- Add missing basic test cases and cover the previous scenario

More tests will be added in a follow-up PR, but this was necessary to prepare the files to receive those upcoming tests.

### DIFF
--- a/packages/react-draggable-dialog/src/components/DraggableDialog/DraggableDialog.test.tsx
+++ b/packages/react-draggable-dialog/src/components/DraggableDialog/DraggableDialog.test.tsx
@@ -4,10 +4,13 @@ import { DraggableDialog } from './DraggableDialog';
 
 describe('DraggableDialog', () => {
   it('should render', () => {
-    render(
-      <DraggableDialog open>
-        <div>Content</div>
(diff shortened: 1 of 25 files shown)


packages/react-draggable-dialog/src/components/DraggableDialog/DraggableDialog.test.tsx
```diff
@@ -4,10 +4,13 @@ import { DraggableDialog } from './DraggableDialog';
 
 describe('DraggableDialog', () => {
   it('should render', () => {
-    render(
-      <DraggableDialog open>
-        <div>Content</div>
+    const text = 'Context';
+    const { getByText } = render(
+      <DraggableDialog>
+        <div>{text}</div>
```
ValentinaKozlova, line 10: Is it required to have `div` inside of Dialog?

ValentinaKozlova (approved):

PR #109, opened 2024-02-19
test(react-draggable-dialog): add test cases for DraggableDialog

marcosmoura:
Add tests for main DraggableDialog component


packages/react-draggable-dialog/src/components/DraggableDialog/DraggableDialog.test.tsx
```diff
@@ -1,16 +1,183 @@
...
-      <DraggableDialog>
+      <DraggableDialog open>
         <div>{text}</div>
       </DraggableDialog>
     );
 
     expect(() => getByText(text)).toBeTruthy();
   });
+
+  it('should return default values', () => {
```
Hotell, line 31: wondering why are we testing internal hook ( not part of public api ) instead testing the component ?

packages/react-draggable-dialog/src/components/DraggableDialog/DraggableDialog.test.tsx
```diff
@@ -1,16 +1,183 @@
 import * as React from 'react';
-import { render } from '@testing-library/react';
+import { act, render, renderHook } from '@testing-library/react';
+
 import { DraggableDialog } from './DraggableDialog';
+import { useDraggableDialog } from './useDraggableDialog';
+import { DragEndEvent } from '@dnd-kit/core';
+
+jest.mock('@fluentui/react-components', () => ({
```
Hotell, line 8: is this mock necessary (mocks most of the time show/expose not the best API/red flags) ? cant we use use`expect.any<String>` in order to test those dictionary outputs of the hook ?

packages/react-draggable-dialog/src/components/DraggableDialog/DraggableDialog.test.tsx
marcosmoura: This component only renders child elements and holds state as Context to provide data down the tree. Since we separate render, state and style into files, the idea of testing the state hook here is to make sure we have correct state passed down as well as reacting to prop changes in the component (since the hook is tied with props).
What do you think about it? Do we have any other approach I can apply here?

packages/react-draggable-dialog/src/components/DraggableDialog/DraggableDialog.test.tsx
marcosmoura: I'm not sure if that is necessary and if we can have an alternative approach. 
The gymnastics here is to make sure that the `useId` hook always returns the same value for the same component and that the value is predictable. This way the tests are both not flaky and the ID returned stays the same, even if useId changes its implementation internally.
Is there any alternative way to achieve the same in this case?

Hotell: yes I mentioned it in my comment. 

regarding flaky - testing internals of useId is probably flaky on itself/codesmell.

`cant we use useexpect.any<String> in order to test those dictionary outputs of the hook ?`

https://jestjs.io/docs/expect#expectanyconstructor

packages/react-draggable-dialog/src/components/DraggableDialog/DraggableDialog.test.tsx
Hotell: if you really need to test the implementation details ( the hook in this case ), I'd move it to related file to the hook implementation -> `useDraggableDialog.test.ts` 

I'd expect DraggableDialog.test to contain test for DraggableDialog component behaviours, not internal hook tests if that makes sense

Hotell (approved):

PR #112, opened 2024-02-20
chore(react-chat): change ChatMessage/ChatMyMessage to new slot api

YuanboXue-Amber:
as title


bsunderhus (approved):

packages/react-chat/src/components/ChatMessage/renderChatMessage.tsx
```diff
@@ -1,36 +1,37 @@
-import * as React from 'react';
-import { getSlots } from '@fluentui/react-components';
+/** @jsxRuntime classic */
+/** @jsx createElement */
+
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+import { createElement } from '@fluentui/react-jsx-runtime';
```
layershifter, line 5:
```
/** @jsxRuntime automatic */
/** @jsxImportSource @fluentui/react-jsx-runtime */
```

Why don't use proper jsxRuntime?

PR #117, opened 2024-02-28
chore(react-tree-grid): minor adjustments on stories

bsunderhus:
Modifies stories to be more accessible:

# Todo story

1. Removes story entirely, as there are some minor issues with it on a11y cases

# Email Inbox

1. Adds custom content with different amount of e-mails on inbox

# Meet App

1. Ensures weekdays correspond with the actual date
2. Adds label to interactive tags


packages/react-tree-grid/stories/EmailInbox.stories.tsx
```diff
@@ -25,10 +25,11 @@ import {
   MenuPopover,
   MenuList,
   MenuItem,
+  Avatar,
 } from '@fluentui/react-components';
 import {
-  CaretRight16Filled,
-  CaretDown16Filled,
+  ChevronRight16Regular,
+  ChevronDown16Regular,
```
ling1726, line 32: don't use sized icons

ling1726 (approved):

PR #119, opened 2024-02-28
feat(houdini-utils): initial implementation

annabratseiko:
Add houdini-utils package to be used together with Liveness indicator components in TMP.


Hotell:
thanks for PR!


the process of creating new packages in repo is following:

- bootstrap the package -> create PR -> merge
- add implementation etc -> different PR

Why?:

- granular PRs === easy to review 
- infra person have no context about implementation of package A, so approving a PR with that in mind is not something we would like to do


also pipeline is failing 
image

Hotell (changes requested):
please dissect these changes in 2 PRs

- scaffold 
- implementation

ty

packages/houdini-utils/src/util/addModule.ts
```diff
@@ -0,0 +1,90 @@
...
+ * When passing a reference pass the class itself, not an instance of the class. E.g.:
+ *
+ *   `addModule("myWorklet", MyHoudiniClass)`
+ *  *  *
+ * @param name - Name of the paint worklet (e.g., "flair")
+ * @param classRefOrString - A class reference or a string representation of the class
+ * @param className - Optional class name. When not passed the class's name will be derived from the class ref or string.
+ * @returns void
+ */
+export const addModule: AddPaintWorkletModuleFn = (
```
ling1726, line 80: As discussed let's update this to work with URLs, we can export the stringify and blobify utilities separately

Hotell:
can we update PR title after the changes ?

`add houdini-utils package` -> `feat(houdini-utils): initial implementation`

packages/houdini-utils/src/paint/fallback/keyframe.ts
```diff
@@ -0,0 +1,109 @@
...
+    const duration = durations[animationIndex] ?? 0;
+    const animDuration = delay + duration;
+
+    if (animDuration > overallDuration) {
+      overallDuration = animDuration;
+    }
+
+    const timingFunction = timingFunctions[animationIndex] ?? 'linear';
+
+    const styles = getComputedStyle(target);
```
spmonahan, line 55: `target` doesn't change with each iteration and `getComputedStyle()` is a relatively expensive call. We should be able to hoist this up outside the loop to improve perf.

packages/houdini-utils/src/util/blobify.ts
```diff
@@ -0,0 +1,51 @@
...
+
+  let str = clsStr;
+  str += '\n';
+  str += `registerPaint('${name}', ${clsName})`;
+
+  return str;
+};
+
+/**
+ * Creates a Blob from a paint worklet.
```
spmonahan, line 38: Since we know this is likely to have issues with CSPs we should add a note about the limitations of this function.

packages/houdini-utils/src/util/featureDetect.ts
```diff
@@ -0,0 +1,94 @@
...
+let canUseHoudiniCache: boolean | undefined = undefined;
+/**
+ * A more robust test for CSS Houdini APIs. This function also verifies that a simple
+ * PaintWorklet can be loaded.
+ *
+ * Note that this check will add a PaintWorklet as a blob URL, apps with CSP policies will always fail this check.
+ * @see hasHoudini
+ * @returns `Promise<true>` if the browser supports necessary Houdini APIs, `Promise<false>` otherwise.
+ */
+export const canUseHoudini: AsyncFeatureDetectFn = async () => {
```
spmonahan, line 67: I'd like to eliminate this function. It's async which introduces a race condition when loading the PaintWorklet and will always fail for apps with modest CSP settings -- preventing them from using the Houdini feature they might otherwise be able to use.

Instead I think we should rely on the synchronous `hasHoundini()` as a baseline test of whether the necessary APIs exist and rely on the result of `addModule()` to potentially use the fallback in the case a Houdini-supporting browser fails to load the `PaintWorklet` for some reason.

spmonahan (approved):
Left a few comments but nothing blocking an initial implementation imo.

ling1726 (approved):

PR #120, opened 2024-02-29
create houdini-utils package

annabratseiko:
Create an empty houdini-utils package.


ling1726 (approved):

packages/houdini-utils/src/index.test.ts
```diff
@@ -0,0 +1,5 @@
+describe('init', () => {
```
Hotell, line 1: this shouldn't be necessary for package scaffold - addressed https://github.com/microsoft/fluentui-contrib/pull/121

ling1726: cool waiting for the merge :)

PR #127, opened 2024-03-06
test(react-draggable-dialog): remove forced mocks

marcosmoura:
A quick fix removing forced mocks for the entire react-components package

### DIFF
--- a/packages/react-draggable-dialog/src/components/DraggableDialog/DraggableDialog.test.tsx
+++ b/packages/react-draggable-dialog/src/components/DraggableDialog/DraggableDialog.test.tsx
@@ -5,18 +5,9 @@ import { DraggableDialog } from './DraggableDialog';
 import { useDraggableDialog } from './useDraggableDialog';
 import { DragEndEvent } from '@dnd-kit/core';
 
-jest.mock('@fluentui/react-components', () => ({
-  ...jest.requireActual('@fluentui/react-components'),
-  useId: jest.fn().mockReturnValue('draggable-dialog'),
-}));
-
 const dialogChild = React.createElement('div', null, 'Dialog Child');
 
 describe('DraggableDialog', () => {
-  afterAll(() => {
-    jest.resetAllMocks();
-  });
-
   it('should render', () => {
     const text = 'Context';
     const { getByText } = render(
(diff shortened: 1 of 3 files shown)


packages/react-draggable-dialog/src/components/DraggableDialogHandle/DraggableDialogHandle.test.tsx
```diff
@@ -7,19 +7,15 @@ import * as contexts from '../../contexts/DraggableDialogContext';
...
-
-jest.mock('@dnd-kit/core', () => {
-  return {
-    __esModule: true,
-    ...jest.requireActual('@dnd-kit/core'),
-  };
-});
+jest.mock('../../contexts/DraggableDialogContext', () => ({
+  __esModule: true,
+  ...jest.requireActual('../../contexts/DraggableDialogContext'),
```
Hotell, line 12: why is this needed ? mocking whole module and using real implementation ?

packages/react-draggable-dialog/src/components/DraggableDialog/DraggableDialog.test.tsx
```diff
@@ -5,18 +5,9 @@ import { DraggableDialog } from './DraggableDialog';
 import { useDraggableDialog } from './useDraggableDialog';
 import { DragEndEvent } from '@dnd-kit/core';
 
-jest.mock('@fluentui/react-components', () => ({
```
Hotell, line 8: noice

ValentinaKozlova (approved):

packages/react-draggable-dialog/src/components/DraggableDialogHandle/DraggableDialogHandle.test.tsx
marcosmoura: This is needed to be able to Spy on functions that come from that module. Whatever Jest does, hoisting the variables when compiling the test files, doesn't work with ES Modules.

Hotell: if we really sure we need such amount of mocks etc, we should be explicit.

Take a look at this https://github.com/microsoft/fluentui-contrib/pull/129/commits/c68318d055eee5b00fa2ac9c46ffff99668d3d68

we can talk about it more offline

marcosmoura: This was added before the availability of the E2E tests. I'm cancelling this PR and I'll those tests in a separate PR in favor of the E2E tests.

PR #138, opened 2024-03-15
feat: set initial position and boundary elements

marcosmoura:
- Features:
  - Ability to set the initial position for the Dialog
  - Ability to set a boundary element that traps the Dialog while dragging
- Remove unnecessary tests
- Code cleanup


ValentinaKozlova (approved):
Discussed offline

Issue #141, opened 2024-03-19
bug( react-data-grid-react-window-grid) columnWidth not called on rerenderer


FengQiMS:
There is cache in 'react-window', similar issue [436](https://github.com/bvaughn/react-window/issues/436)

layershifter:
Fixed in #142

PR #142, opened 2024-03-19
minor(react-data-grid-react-window-grid): bug fix on columnWidth not been called when rerendered

FengQiMS:
Fix issue #141, There is cache from 'react-window' for performance improvement reason. here is the doc how to force update:

> `resetAfterColumnIndex(index: number, shouldForceUpdate: boolean = true): void`
> VariableSizeGrid caches offsets and measurements for each column index for performance purposes. This method clears that 
> cached data for all columns after (and including) the specified index. It should be called whenever a column's width changes. (Note that this is not a typical occurrance.)
>
> By default the grid will automatically re-render after the index is reset. If you would like to delay this re-render until e.g. a state update has completed in the parent component, specify a value off false for the second, optional parameter.

Before fix the bug behaves:
beforefixwidth
After fix:
afterfixwidth

### DIFF
--- a/packages/react-data-grid-react-window-grid/src/components/DataGrid/useDataGrid.ts
+++ b/packages/react-data-grid-react-window-grid/src/components/DataGrid/useDataGrid.ts
@@ -1,7 +1,5 @@
 import * as React from 'react';
-import {
-  useDataGrid_unstable as useBaseState,
-} from '@fluentui/react-components';
+import { useDataGrid_unstable as useBaseState } from '@fluentui/react-components';
 import { useFluent, useScrollbarWidth } from '@fluentui/react-components';
 import { DataGridProps, DataGridState } from './DataGrid.types';
 import { VariableSizeGrid, VariableSizeList } from 'react-window';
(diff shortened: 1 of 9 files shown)


packages/react-data-grid-react-window-grid/src/components/DataGrid/useDataGrid.ts
```diff
@@ -33,8 +31,14 @@ export const useDataGrid_unstable = (
...
-  return useBaseState(
+  const baseState = useBaseState(
     { ...props, 'aria-rowcount': props.items.length, containerWidthOffset },
     ref
   );
+
+  return {
+    ...baseState,
+    bodyRef: props.bodyRef ?? React.useRef<VariableSizeGrid>(null),
+    headerRef: props.headerRef ?? React.useRef<VariableSizeList>(null),
```
layershifter, line 42: Hooks can't be called conditionally

FengQiMS: Thanks for the tips, will update it with useMergedRefs

packages/react-data-grid-react-window-grid/stories/DataGrid/VirtualizedDataGrid.stories.tsx
```diff
@@ -100,42 +100,61 @@ export const VirtualizedDataGrid: React.FunctionComponent = () => {
   );
   const items = generateTableArrays(1000, 50);
   const styles = useStyles();
-
+  const bodyRef = React.useRef<VariableSizeGrid>(null);
+  const headerRef = React.useRef<VariableSizeList>(null);
+  const [width, setWidth] = React.useState(200);
+  const columnWidth = React.useCallback(
+    (index: number) => (index == 0 ? 200 : width),
```
stkevintan, line 107: `index === 0`

stkevintan (approved):

PR #144, opened 2024-03-21
test(react-draggable-dialog): add config for E2E tests

marcosmoura:
Add E2E config files that generator outputs


packages/react-draggable-dialog/playwright.config.ts
```diff
@@ -0,0 +1,55 @@
+import { defineConfig, devices } from '@playwright/experimental-ct-react';
+
+/**
+ * Read environment variables from file.
+ * https://github.com/motdotla/dotenv
+ */
+// require('dotenv').config();
```
ValentinaKozlova, line 7: do you need this comment here?

ValentinaKozlova (approved):

packages/react-draggable-dialog/playwright.config.ts
Hotell: that's added by generator

PR #145, opened 2024-03-21
chore: scaffold react-themeless-provider package

spmonahan:
Scaffolds a `ThemelessFluentProvider` component per [FluentUI#30241](https://github.com/microsoft/fluentui/issues/30241)


packages/react-themeless-provider/tsconfig.storybook.json
```diff
@@ -0,0 +1,31 @@
...
+    "src/**/*.test.ts",
+    "src/**/*.spec.js",
+    "src/**/*.test.js",
+    "src/**/*.spec.tsx",
+    "src/**/*.test.tsx",
+    "src/**/*.spec.jsx",
+    "src/**/*.test.js"
+  ],
+  "include": [
+    "src/**/*.stories.ts",
```
ling1726, line 23: I think the nx upgrade changed the setup of storybook projects. This tsconfig will not give you any typechecking in your stories since `/stories` is not under `src` - to align with the core repo

There is also another tsconfig in the `.storybook` folder which was used in the last major version of nx which is no longer the case.

@Hotell what would the ideal way forward be here? should the current generator be updated to override the nx generated tsconfig?

spmonahan: @ling1726  and @Hotell what do we need to do to get resolution on this? I have [a follow up](https://github.com/microsoft/fluentui-contrib/pull/149) that depends on this so I'd like to get it merged.

LMK what you need from me -- happy to do the work.

Hotell: https://github.com/microsoft/fluentui-contrib/pull/151

Hotell: merged, sorry for issues caused and thanks @ling1726 for eagles eye !

PR #153, opened 2024-04-02
chore: scaffold react-themeless-provider package

spmonahan:
Scaffolds a `ThemelessFluentProvider` component per [FluentUI#30241](https://github.com/microsoft/fluentui/issues/30241)

Previously this was in #145 but there were issues with the generator scripts so I spun up a new PR.


packages/react-themeless-provider/stories/ThemelessFluentProvider/index.stories.tsx
```diff
@@ -0,0 +1,10 @@
+// import * as React from 'react';
```
Hotell, line 1: lets not ship cruft
```suggestion
```

Hotell (approved):

PR #155, opened 2024-04-03
Nx plugin/fix nx warnings and flaky tests

Hotell:


packages/nx-plugin/src/generators/component-playwright-spec/generator.spec.ts
```diff
@@ -17,11 +20,12 @@ describe('playwright-component-configuration generator', () => {
 
   beforeEach(async () => {
     tree = createTreeWithEmptyWorkspace();
+    setupWorkspaceDependencies(tree);
     createCodeowners(tree);
 
     await libraryGenerator(tree, { name: 'hello', owner: '@MrWick' });
     await componentGenerator(tree, { name: 'hello', componentName: 'Hello' });
-  }, 10000);
```
Hotell, line 24: 🚨

PR #158, opened 2024-04-03
fix(nx-plugin): make `component` generator result pass ci checks

spmonahan:
## Previous Behavior

When following the [contributing guidelines](https://github.com/microsoft/fluentui-contrib/blob/main/CONTRIBUTING.md#creating-a-new-package) a freshly scaffolded package would not pass the build.

before

## Current Behavior

Generator templates have been updated so a freshly minted package passes the build.

after

### DIFF
--- a/packages/nx-plugin/src/generators/component/generator.ts
+++ b/packages/nx-plugin/src/generators/component/generator.ts
@@ -1,9 +1,4 @@
-import {
-  formatFiles,
-  generateFiles,
-  Tree,
-  workspaceRoot,
-} from '@nx/devkit';
+import { formatFiles, generateFiles, Tree, workspaceRoot } from '@nx/devkit';
 import * as path from 'path';
 import { ComponentGeneratorSchema } from './schema';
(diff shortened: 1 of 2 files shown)


Hotell:
extra kudos for your shell prompt ! 
image

Hotell (approved):
ty!

packages/nx-plugin/src/generators/component/generator.ts
```diff
@@ -15,14 +10,20 @@ export default async function (tree: Tree, options: ComponentGeneratorSchema) {
...
+    npmScope,
+  });
 
   let main = tree.read(paths.main)?.toString();
   if (!main) {
     throw new Error('No index.ts main entrypoint');
   }
 
-  main = `export * from './components/${componentName}';` + '\n' + main;
+  main =
```
Hotell, line 23: we could also run `eslint --fix` which would produce the same output, which for more complicated use-case is the way to go. for this use case this is good enough and won't slow down generator execution by spawning another process

PR #165, opened 2024-04-08
feat(react-draggable-dialog): prepare for first release

marcosmoura:
Remove the private flag from package to enable the release of the Draggable Dialog


packages/react-draggable-dialog/package.json
```diff
@@ -1,7 +1,6 @@
 {
   "name": "@fluentui-contrib/react-draggable-dialog",
   "version": "0.0.1",
-  "private": true,
```
Hotell, line 4: let's also add this to enforce zero based semver 

```json
"beachball": {
    "disallowedChangeTypes": [
      "major",
      "prerelease"
    ]
  },
```

Hotell (approved):
I also updated title to be self explanatory, hope you dont mind 🙌

PR #167, opened 2024-04-09
chore: dedupe jest and bump babel/jest related deps

Hotell:
- dedupes `jest` so we have only 1 instance
- bumps jest/babel and related packages to mitigate dupes
- draggable-dialog test started failing after this dep cleanup/version bump - so I applied my suggestions from https://github.com/microsoft/fluentui-contrib/pull/129 which makes the test pass 

### Depends on:
- [x] https://github.com/microsoft/fluentui-contrib/pull/164
- Closes https://github.com/microsoft/fluentui-contrib/pull/129


marcosmoura (approved):

marcosmoura:
I tried applying what was mentioned in the #129 but it wasn't working. Glad you found the culprit (jest and its deps)! 👍🏼

PR #149, opened 2024-03-22
Add `ThemelessFluentProvider` implementation

spmonahan:
Implements `ThemelessFluentProvider` component per [FluentUI#30241](https://github.com/microsoft/fluentui/issues/30241).

Adds:

1. Implementation
2. Documentation/Storybook example
3. Tests

Depends on https://github.com/microsoft/fluentui-contrib/pull/153

PR #179, opened 2024-05-21
feat: implement verify integrity check

Hotell:
1. Removes invalid changefiles introduced
  - https://github.com/microsoft/fluentui-contrib/pull/143/files#diff-935baad70042ef84471b919f884a2f7b73ef42ddfd003dc059c50310b819bebe
  - https://github.com/microsoft/fluentui-contrib/pull/149/files#diff-0862bad8eb7dca9166db074d664922457daf1349e8084c75c1e29f93b6ce097a 
  - image

2. introduces verify-integrity target within root repo which will be executed only if root repo files are affected

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "fluentui-contrib",
   "private": true,
-  "version": "0.1.0",
+  "version": "0.0.0",
   "license": "MIT",
   "scripts": {
     "build": "nx affected --target=build",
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "fluentui-contrib",
   "private": true,
-  "version": "0.1.0",
```
Hotell, line 4: undo previous unintended change

tudorpopams (approved):

PR #181, opened 2024-05-23
chore: update eslint to disallow globals

spmonahan:
Accessing global objects can cause problems in multi-window applications because we may be accessing a global object in the wrong `window` context. 

This change enables the `eslint` `no-restricted-globals` rule to disallow global access.

For more details see the corresponding PR in Fluent UI: https://github.com/microsoft/fluentui/pull/30967

This change does not address problems in `houdini-utils` specifically as those will be addressed separately.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -6,7 +6,7 @@
     "no-restricted-globals": [
       "error",
       "window",
-      "document"
+      "document",
       "customElements",
       "devicePixelRatio",
       "location",
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -2,6 +2,42 @@
   "root": true,
   "ignorePatterns": ["**/*"],
   "plugins": ["@nx", "import", "@rnx-kit"],
+  "rules": {
+    "no-restricted-globals": [
+      "error",
+      {
+        "name": "window",
+        "message": "Get a reference to `window` from `useFluent()`."
+      },
+      {
+        "name": "document",
+        "message": "Get a reference to `document` from `useFluent()`."
+      },
+      "customElements",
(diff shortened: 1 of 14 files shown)


packages/houdini-utils/src/paint/fallback/util/wrapper.ts
```diff
@@ -1,7 +1,11 @@
 const createWrapper = (id: string): HTMLElement => {
+  // TODO: fix global
```
Hotell, line 2: can we create issues and ref it here to have clear actionable for team owning this package ?

.eslintrc.json
```diff
@@ -2,6 +2,42 @@
   "root": true,
   "ignorePatterns": ["**/*"],
   "plugins": ["@nx", "import", "@rnx-kit"],
+  "rules": {
+    "no-restricted-globals": [
+      "error",
+      {
+        "name": "window",
+        "message": "Get a reference to `window` from `useFluent()`."
```
Hotell, line 10: not sure if this is the right configuration file as not all packages in this repo are tightly coupled to fluent => users might get false positives with this.

packages/houdini-utils/src/paint/fallback/util/wrapper.ts
spmonahan: Added #183, will push a change updating the comments to reference it.

.eslintrc.json
spmonahan: Switched to defaults for now.

Hotell (approved):
TY!

ling1726 (approved):

PR #189, opened 2024-06-28
Proposal: Introduce react-fluentui-examples project

george-cz:
# Motivation

Historically, in the effort of supporting our partners, we've built a _lot_ of examples in Stackblitz displaying how to use our components in a particual use case. These use cases don't usually (never?) make it back to the original docsite, even though others could potentially benefit from them too. The reasons for this are multiple:
 - these examples are not generic enough
 - these examples don't demonstrate the component's prop directly, but **build on top of the component** to get a closer idea how to productionalize the component
 - these examples may combine multiple components
 - the doc examples have to be carefully curated, straightforward and to the point of what part of the API are they demonstrating

The problem with the current approach is that these carefully crafted examples end in up `/dev/null` eventually because we forget about them, or don't about their existence in the first place. Having one place where we could store these, link to them and let the user quickly run them and play with them would be benefitial.

# Other attempts
The [Fluent UI v0](https://fluentsite.z22.web.core.windows.net/0.66.2/prototype-chat-messages) already has a concept of this, they call it _Prototypes_. This means that the idea behind this concept has been validated in the past, has proven useful and we should strive to provide a similar solution to our V9 docs.

# How this works

The main ideas behind this PR are:
- We have a place where we can store these more complex examples
- Github repo is easy to share and Stackblitz has support to load code directly from here (Stackblitz provides a nice button!)

[Visit the project here](https://github.com/microsoft/fluentui-contrib/tree/jirivyhnalek/react-fluentui-examples/packages/react-fluentui-examples)

# Other considerations
- We might want to consider tighter integration with the official doc site in the future
- Would be nice to have a script that bootstraps the example (something I can work on if this gains some traction)
- Granularity level for the examples: should the `react-fluentui-examples` be just one package and examples just storybook? Storybook has limitations, the current approach (each example is its own package) is similar to maintain, use in Stackblitz and provides more flexibility though.


ling1726:
I think we should consider a way to get these examples served by the official docsite. It would be good to have a single source of truth and give some more weight to the contrib packages

packages/react-fluentui-examples/List/list-active-element/src/main.tsx
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+import ReactDOM from 'react-dom/client';
+import { FluentProvider, webLightTheme } from '@fluentui/react-components';
+
+import Example from './List-Active-Element.tsx';
+
+ReactDOM.createRoot(document.getElementById('root')!).render(
+  <React.StrictMode>
+    <FluentProvider theme={webLightTheme}>
+      <Example />
```
ling1726, line 10: Let's add some kind of mounting/routing support - Otherwise there's too much barrier for someone to contribute the 2nd,3rd examples because then they would need to build that

packages/react-fluentui-examples/List/list-active-element/package.json
```diff
@@ -0,0 +1,29 @@
+{
+  "name": "react-fluentui-examples-list-active-element",
```
ling1726, line 2: There's no build infra in the contrib repro to handle application type projects, as all the infra is built around publishable packages. The task runner (NX) isn't even integrated to this package and it maintains its own dependencies.

I'd suggest to move it out to another `/apps` folder on the root to keep the separation from the publishable packages

packages/react-fluentui-examples/List/list-active-element/src/List-Active-Element.tsx
```diff
@@ -0,0 +1,125 @@
...
+            key={name}
+            value={name}
+            className={mergeClasses(
+              classes.item,
+              activeItem === name && classes.itemActive
+            )}
+            onAction={onAction}
+            data-value={name}
+            aria-label={name}
+            onFocus={onFocus}
```
jurokapsiar, line 88: please test this with JAWS browser mode as that doesn't send focus events as you navigate.
Also, try with at least windows touch screen device if you are still able to change the value with just tap

packages/react-fluentui-examples/List/list-active-element/src/List-Active-Element.tsx
```diff
@@ -0,0 +1,125 @@
...
+  return (
+    <div>
+      <List navigationMode="composite">
+        {items.map(({ name, avatar }) => (
+          <ListItem
+            key={name}
+            value={name}
+            className={mergeClasses(
+              classes.item,
+              activeItem === name && classes.itemActive
```
jurokapsiar, line 83: should there also be aria-selected or some other aria prop indicating the selection state?

Hotell:
AFAIR we have react-recipes in core repo for this exact purpose https://react.fluentui.dev/?path=/docs/concepts-recipes-media-object--docs

@sopranopillow can provide more details if needed.

sopranopillow:
@Hotell is correct, that is the purpose of recipes 😄

PR #190, opened 2024-07-01
Azure Theme V9 Object + Stories

adr1-jpg:
This PR adds the implementation logic for azure-theme package in V9 components. 
The style object includes a brand ramp and neutral colors for Azure. 
Storybook includes light and dark themes, with many Fluent 2 controls nested in. 

image


Jacqueline-ms (approved):
Adrianna has implemented the style objects for light and dark theme, while I have added a storybook so stakeholders can see how the object looks applied to Fluent 2 controls. After many iterations, this looks good to go!

packages/azure-theme/package.json
```diff
@@ -2,6 +2,7 @@
   "name": "@fluentui-contrib/azure-theme",
   "version": "0.0.1",
   "dependencies": {
+    "@fluentui/react-icons": ">=2.0.204 <3.0.0",
```
Hotell, line 5: had offline convo with @Jacqueline-ms 

in order to start releasing this package:
- remove `private'  field completely
- run `yarn change` to generate changefile 
- merge
- we can release

packages/azure-theme/package-lock.json
```diff
@@ -0,0 +1,1741 @@
+{
```
Hotell, line 1: hmm this shouldn't be here. please as follow-up remove this lock file.

PR #194, opened 2024-07-03
feat(react-resize-handle): Allow relative panel adjustments

george-cz:
This PR adds possibility to use relative sizes in combination with resizable panels. Only the difference from the default dimension is measured, which, when plugged in the `calc` function, results in only relative adjustments. This still fixes 2 issues:

- issues where the panel is first rendered with 0 width, as that will immediately measure it and set as fixed
- Issue where the layout would become irresponsive as a result of plugging this in, as all of the sizes would be in px.

### DIFF
--- a/packages/react-resize-handle/src/hooks/useKeyboardHandler.ts
+++ b/packages/react-resize-handle/src/hooks/useKeyboardHandler.ts
@@ -5,8 +5,8 @@ import { elementDimension } from '../utils/index';
 
 export type UseKeyboardHandlerOptions = {
   onValueChange: (value: number) => void;
-  elementRef: React.RefObject<HTMLElement>;
   growDirection: GrowDirection;
+  currentValue: React.RefObject<number>;
 };
 
 const DEFAULT_STEP = 20;
@@ -44,19 +44,21 @@ function isSupportedKey(
 }
 
 export const useKeyboardHandler = (options: UseKeyboardHandlerOptions) => {
-  const { elementRef, onValueChange, growDirection } = options;
+  const { onValueChange, growDirection, currentValue } = options;
   const { dir } = useFluent();
 
   const onKeyDown = useEventCallback((event: KeyboardEvent) => {
-    let newValue = elementDimension(elementRef.current, growDirection);
-
-    if (isSupportedKey(growDirection, event.key)) {
-      const multiplier = multipliers[growDirection][event.key] ?? 1;
-      const directionMultiplier =
-        dir === 'rtl' && ['start', 'end'].includes(growDirection) ? -1 : 1;
-
-      newValue += multiplier * DEFAULT_STEP * directionMultiplier;
+    if (!isSupportedKey(growDirection, event.key)) {
+      return;
     }
+
+    let newValue = currentValue.current || 0;
+
+    const multiplier = multipliers[growDirection][event.key] ?? 1;
+    const directionMultiplier =
+      dir === 'rtl' && ['start', 'end'].includes(growDirection) ? -1 : 1;
+
+    newValue += multiplier * DEFAULT_STEP * directionMultiplier;
 
     onValueChange(Math.round(newValue));
   });
@@ -67,6 +69,7 @@ export const useKeyboardHandler = (options: UseKeyboardHandlerOptions) => {
     },
     [onKeyDown]
   );
+
   const detachHandlers = React.useCallback(
     (node: HTMLElement) => {
       node.removeEventListener('keydown', onKeyDown);
--- a/packages/react-resize-handle/src/hooks/useMouseHandler.ts
+++ b/packages/react-resize-handle/src/hooks/useMouseHandler.ts
@@ -7,16 +7,15 @@ import {
 } from '@fluentui/react-utilities';
 import * as React from 'react';
 import { GrowDirection } from '../types';
-import { elementDimension } from '../utils';
 
 export type UseMouseHandlerParams = {
   onDown?: (event: NativeTouchOrMouseEvent) => void;
   onMove?: (event: NativeTouchOrMouseEvent) => void;
-  elementRef: React.RefObject<HTMLElement>;
   growDirection: GrowDirection;
   onValueChange: (value: number) => void;
   onDragEnd?: (e: NativeTouchOrMouseEvent) => void;
   onDragStart?: (e: NativeTouchOrMouseEvent) => void;
+  currentValue: React.RefObject<number>;
 };
 
 export function useMouseHandler(params: UseMouseHandlerParams) {
@@ -24,11 +23,9 @@ export function useMouseHandler(params: UseMouseHandlerParams) {
   const targetWindow = targetDocument?.defaultView;
 
   const dragStartOriginCoords = React.useRef({ clientX: 0, clientY: 0 });
-  const { growDirection, onValueChange, elementRef } = params;
+  const { growDirection, onValueChange, currentValue } = params;
 
-  const initialElementSize = React.useRef(
-    elementDimension(elementRef.current, growDirection)
-  );
+  const startValue = React.useRef(0);
 
   const recalculatePosition = useEventCallback(
     (event: NativeTouchOrMouseEvent) => {
@@ -38,7 +35,7 @@ export function useMouseHandler(params: UseMouseHandlerParams) {
         clientY - dragStartOriginCoords.current.clientY,
       ];
 
-      let newValue = initialElementSize.current;
+      let newValue = startValue.current || 0;
 
       switch (growDirection) {
         case 'end':
@@ -56,15 +53,6 @@ export function useMouseHandler(params: UseMouseHandlerParams) {
       }
 
       onValueChange(Math.round(newValue));
-
-      // If, after resize, the element size is different than the value we set, that we have reached the boundary
-      // and the element size is controlled by something else (minmax, clamp, max, min css functions etc.)
-      // In this case, we need to update the value to the actual element size so that the css var and a11y props
-      // reflect the reality.
-      const elSize = elementDimension(elementRef.current, growDirection);
-      if (elSize !== newValue) {
-        onValueChange(elSize);
-      }
     }
   );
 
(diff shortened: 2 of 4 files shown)


packages/react-resize-handle/src/hooks/useResizeHandle.ts
```diff
@@ -40,6 +40,14 @@ export type UseResizeHandleParams = {
    * Use this for localization.
    */
   getA11ValueText?: (value: number) => string;
+
+  /**
+   * Only measure relative change in size, useful to use with CSS calc() function.
+   * Example: clamp(60px, calc(20% + var(--nav-size)), 40%)
```
george-cz, line 46: I think this should be `+`, fixing

packages/react-resize-handle/src/hooks/useResizeHandle.ts
```diff
@@ -51,7 +59,9 @@ export const useResizeHandle = (params: UseResizeHandleParams) => {
     onChange,
     onDragStart,
     onDragEnd,
-    getA11ValueText = (value) => `${value.toFixed(0)}px`,
+    getA11ValueText = (value) =>
+      value === UNMEASURED ? 'Unknown' : `${value.toFixed(0)}px`,
```
ling1726, line 63: Let's just return undefined instead of 'Unknown' - since we don't want a default un localized screen reader message.

Also in the documentation there is no mention that a value of `-1` is considered unmeasured - let's reflect that. I think it should be enough to add a11y text to the story example.

packages/react-resize-handle/src/hooks/useResizeHandle.ts
```diff
@@ -61,18 +71,23 @@ export const useResizeHandle = (params: UseResizeHandleParams) => {
...
-      `${currentValue.current}px`
-    );
+    let a11yValue = getA11ValueText(currentValue.current);
+
+    // If relative mode is enabled, we actually have to measure the element,
+    // because the currentValue is just the px offset.
+    if (relative) {
+      const sizeInPx = elementDimension(elementRef.current, growDirection);
+      a11yValue = getA11ValueText(sizeInPx);
+    }
```
ling1726, line 81:
```suggestion
    const a11yValue = relative
      ? // If relative mode is enabled, we actually have to measure the element,
        // because the currentValue is just the px offset.
        getA11ValueText(elementDimension(elementRef.current, growDirection))
      : getA11ValueText(currentValue.current);
```

packages/react-resize-handle/src/hooks/useResizeHandle.ts
```diff
@@ -164,8 +219,8 @@ export const useResizeHandle = (params: UseResizeHandleParams) => {
     (node) => {
       elementRef.current = node;
       if (elementRef.current) {
-        if (currentValue.current === UNMEASURED) {
-          currentValue.current = elementDimension(node, growDirection);
+        if (currentValue.current === UNMEASURED && relative) {
```
ling1726, line 222: let's remove the condition nesting
```ts
      elementRef.current = node;
      if (
        elementRef.current &&
        currentValue.current === UNMEASURED &&
        relative
      ) {
        currentValue.current = 0;
        updateElementsAttrs();
      }
```

packages/react-resize-handle/src/hooks/useResizeHandle.ts
```diff
@@ -83,20 +98,52 @@ export const useResizeHandle = (params: UseResizeHandleParams) => {
...
 
   // In case the maxValue or minValue is changed, we need to make sure we are not exceeding the new limits
   React.useEffect(() => {
-    setValue(currentValue.current);
-  }, [maxValue, minValue, updateElementsAttrs]);
+    if (currentValue.current === UNMEASURED) {
+      return;
+    }
+    currentValue.current = clamp(currentValue.current, minValue, maxValue);
+  }, [maxValue, minValue]);
```
ling1726, line 117:
```suggestion
    if (currentValue.current === UNMEASURED || relative) {
      return;
    }
    currentValue.current = clamp(currentValue.current, minValue, maxValue);
  }, [maxValue, minValue, relative]);
```

From what I understand relative should completely ignore min/max values

ling1726: Another thought - how about changing the reference of the `clamp` in the entire function to do nothing when `relative` is set? that way you won't need to worry about conditional logic for clamps.... as I understand clamp should never happen with relative

packages/react-resize-handle/src/hooks/useMouseHandler.ts
```diff
@@ -38,7 +35,7 @@ export function useMouseHandler(params: UseMouseHandlerParams) {
         clientY - dragStartOriginCoords.current.clientY,
       ];
 
-      let newValue = initialElementSize.current;
+      let newValue = startValue.current || 0;
```
ling1726, line 38:
```suggestion
      let newValue = startValue.current ?? 0;
```

ling1726: actually is the conditional assignment even necessary? `startValue.current` is always a number

packages/react-resize-handle/src/hooks/useMouseHandler.ts
```diff
@@ -88,10 +76,9 @@ export function useMouseHandler(params: UseMouseHandlerParams) {
 
   const onPointerDown = useEventCallback((event: NativeTouchOrMouseEvent) => {
     dragStartOriginCoords.current = getEventClientCoords(event);
-    initialElementSize.current = elementDimension(
-      elementRef.current,
-      growDirection
-    );
+    // As we start dragging, save the current value otherwise the value increases,
+    // the delta compounds and the element grows/shrinks too fast.
+    startValue.current = getCurrentValue() || 0;
```
ling1726, line 81:
```suggestion
    startValue.current = getCurrentValue();
```

packages/react-resize-handle/src/hooks/useKeyboardHandler.ts
```diff
@@ -44,19 +43,21 @@ function isSupportedKey(
...
-    let newValue = elementDimension(elementRef.current, growDirection);
+    if (!isSupportedKey(growDirection, event.key)) {
+      return;
+    }
 
-    if (isSupportedKey(growDirection, event.key)) {
-      const multiplier = multipliers[growDirection][event.key] ?? 1;
-      const directionMultiplier =
-        dir === 'rtl' && ['start', 'end'].includes(growDirection) ? -1 : 1;
+    let newValue = getCurrentValue() || 0;
```
ling1726, line 54:
```suggestion
    let newValue = getCurrentValue();
```

packages/react-resize-handle/src/hooks/useResizeHandle.ts
```diff
@@ -40,6 +40,14 @@ export type UseResizeHandleParams = {
...
    */
   getA11ValueText?: (value: number) => string;
+
+  /**
+   * Only measure relative change in size, useful to use with CSS calc() function.
+   * Example: clamp(60px, calc(20% + var(--nav-size)), 40%)
+   * This will take 20% of the parent size by default, but will allow to resize between 60px and 40%.
+   * Also, the size will still be relative in nature (20% + (X)px).
+   */
+  relative?: boolean;
```
ling1726, line 50: The repo now has support for playwright tests, we should have tests there. In this case let's not block you with testing the entire thing but just the new relative feature

ling1726, line 50: Let's add doc for the min and max value options that they are ignored if the user sets relative

packages/react-resize-handle/src/hooks/useMouseHandler.ts
george-cz: yea, not anymore, removing

packages/react-resize-handle/src/hooks/useResizeHandle.ts
george-cz: I actually wanted to create a follow up ticket on updating this value with the actual element size when the handle gets focused. But until then, I agree undefined will work better.

packages/react-resize-handle/src/hooks/useResizeHandle.ts
george-cz: I ended up creating `clampWithMode` where the last parameter is `relative?: boolean`

packages/react-resize-handle/src/hooks/useResizeHandle.ts
```diff
@@ -83,20 +98,52 @@ export const useResizeHandle = (params: UseResizeHandleParams) => {
...
         currentValue.current = newValue;
         updateElementsAttrs();
+
+        // Measure the size after setting the new value
+        const newSizeInPx = elementDimension(elementRef.current, growDirection);
+
+        // If the size hasn't changed, we need to revert to the old value to keep the state and DOM in sync.
+        // If we don't do this, then the handle might be stuck in a place where small changes
+        // in value don't have any effect.
+        if (newSizeInPx === previousSizeInPx) {
```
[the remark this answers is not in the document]
george-cz: the `newValue` is the newly calculated potential size for the element, based on the mouse movement for example. What happens when the user tries to resize the element above its maximum possible size (lets say the layout only allows it to go 350px, but they move the cursor right 10 px, newValue will be 360px), this would result into us setting the new value larger and the a11y value being wrong (reading 360 instead of 350). 

Previously this would also result in a 10px "buffer" when the user tries to resize again, but I started to actually measure the element every time they go and resize, making sure we always start from the actual value (in the absolute mode). In relative mode, we are also relying on this code here to only update the css value if it actually has any impact on the element size.

Hope this all makes sense :)

packages/react-resize-handle/src/hooks/useResizeHandle.ts
george-cz: your code only calls `updateElementsAttrs()` when all of the conditions are met, in my code we always call it as soon as we receive the element. This is by design and this code change would break the logic.

ling1726: Sorry typo - I just wanted to flatten the condition. I didn't mean to move the function call inside

ling1726 (approved):

packages/react-resize-handle/src/hooks/useResizeHandle.ts
george-cz: We agreed that follow up PR will include tests as this is quite time sensitive right now

PR #196, opened 2024-07-10
Generate empty variant-theme package with generator

wenyaoj2:
Generate empty variant-theme package with generator

### DIFF
--- a/packages/variant-theme/src/components/variantThemeTool/variantThemeTool.styles.ts
+++ b/packages/variant-theme/src/components/variantThemeTool/variantThemeTool.styles.ts
@@ -1,0 +1,22 @@
+import { makeStyles, shorthands, tokens } from '@fluentui/react-components';
+
+export const useStyles = makeStyles({
+  root: {
+    ...shorthands.padding(tokens.spacingHorizontalM),
+    ...shorthands.border(
+      tokens.strokeWidthThin,
+      'solid',
+      tokens.colorNeutralStroke1
+    ),
+    color: tokens.colorNeutralForeground1,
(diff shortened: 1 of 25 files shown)


packages/variant-theme/src/components/variantThemeTool/variantThemeTool.styles.ts
```diff
@@ -0,0 +1,22 @@
+import { makeStyles, shorthands, tokens } from '@fluentui/react-components';
```
Hotell, line 1: when generating new packages it should not contain implementation 

https://github.com/microsoft/fluentui-contrib/blob/main/CONTRIBUTING.md#creating-a-new-package
image


please undo these changes

Hotell (commented):
let's undo the implementation logic of the package part, then we can merge. ty

packages/variant-theme/src/components/variantThemeTool/variantThemeTool.styles.ts
wenyaoj2: Hi @Hotell , I have removed all these codes, please help to take a look. Thanks

packages/variant-theme/stories/variantThemeTool/Default.stories.tsx
Hotell, line 1: please delete

packages/variant-theme/stories/variantThemeTool/index.stories.tsx
Hotell, line 1: please delete

Hotell (approved):

PR #197, opened 2024-07-11
Fix: default scrolling behavior does not work when the gridProps property of the DataGridBody component has an onScroll function

t-yonghu:


t-yonghu:
I need to customize the `onScroll `function through `gridprops`, but customizing `onScroll `will cause inconsistent scrolling of the **header** and **body** of the table. After reading the component source code, I found that **customizing** `onScroll `will override the **default** `onScroll `function

t-yonghu:
If you have any questions, please contact me alias: t-yonghu

stkevintan:
how about we make the default onScroll function as a parameter of the gridprops's onScroll ?. So developer can decide whether composing or replacing the default onScroll behavior?

PR #202, opened 2024-07-22
feat(variant-theme): add algorithm of variant theme generator

wenyaoj2:
This is the algorithm of variant theme generator

Codes in the ThemeDesigner folder are copied from Theme-Designer package in Fluent https://github.com/microsoft/fluentui/tree/master/packages/react-components/theme-designer

ThemeGenerator.ts is the file that needs to be reviewed.  
getVariantTheme() function will return corresponding variant theme according to the theme and VariantThemeType


packages/variant-theme/package.json
```diff
@@ -2,13 +2,17 @@
   "name": "@fluentui-contrib/variant-theme",
   "version": "0.0.1",
   "dependencies": {
-    "@swc/helpers": "~0.5.2"
+    "@swc/helpers": "~0.5.2",
+    "@fluentui/react-components": ">=9.46.3 <10.0.0",
```
Hotell, line 6: react-components should never be a direct dependency of extension, please remove this

packages/variant-theme/package.json
```diff
@@ -2,13 +2,17 @@
...
+    "@fluentui/react-migration-v8-v9": "^9.6.12",
+    "@fluentui/scheme-utilities": "^8.3.46",
+    "@griffel/react": "^1.5.14",
+    "lodash" :"^4.17.21"
```
Hotell, line 10: unused - pls remove / note nowadays there is very little need to use lodash as most functionality is build in into ECMA already. lets not pollute users bundle-sizes with junk

packages/variant-theme/package.json
```diff
@@ -2,13 +2,17 @@
   "name": "@fluentui-contrib/variant-theme",
   "version": "0.0.1",
   "dependencies": {
-    "@swc/helpers": "~0.5.2"
+    "@swc/helpers": "~0.5.2",
+    "@fluentui/react-components": ">=9.46.3 <10.0.0",
+    "@fluentui/react-migration-v8-v9": "^9.6.12",
+    "@fluentui/scheme-utilities": "^8.3.46",
```
Hotell, line 8: what's the reasoning behind those 2 deps ? 

`react-migration-v8-v9` is also very dangerous as it has react-components as dependency ( that's bug on fluent core side IMO ) - lets remove these

packages/variant-theme/src/ThemeDesigner/src/colors/csswg.ts
Hotell, line 1: lets keep the code shallow, having 2 `src/` folders is very confusing

we can keep the copied files within `variant-theme/src/generated/*` to express the scope and intent.

packages/variant-theme/src/ThemeGenerator.ts
```diff
@@ -0,0 +1,426 @@
+/* eslint-disable */
+/* tslint-disable */
+//@ts-nocheck
+import {
+  getColorFromString,
+  IColor,
+  updateA
+} from '@fluentui/react';
+import { BrandVariants, Theme } from '@fluentui/react-components';
+import { VariantThemeType } from '@fluentui/scheme-utilities';
```
Hotell, line 10: always specify kind of import in order to improve tree-shakibility

```suggestion
import type { VariantThemeType } from '@fluentui/scheme-utilities';
```

packages/variant-theme/src/ThemeGenerator.ts
```diff
@@ -0,0 +1,426 @@
+/* eslint-disable */
+/* tslint-disable */
+//@ts-nocheck
+import {
+  getColorFromString,
+  IColor,
+  updateA
+} from '@fluentui/react';
+import { BrandVariants, Theme } from '@fluentui/react-components';
```
Hotell, line 9:
```suggestion
import type { BrandVariants, Theme } from '@fluentui/react-components';
```

packages/variant-theme/src/ThemeGenerator.ts
```diff
@@ -0,0 +1,426 @@
...
+  grey,
+  whiteAlpha,
+  blackAlpha,
+  AlphaColors,
+  type Greys,
+} from '@fluentui/react-migration-v8-v9';
+import * as lodash from 'lodash';
+import { contrast, hex_to_sRGB } from './ThemeDesigner/src/colors';
+
+const standardContrastRatio: number = 4.5;
```
Hotell, line 23: no need for explicit annotation, ts knows this is `number` from the assigned value

```suggestion
const standardContrastRatio = 4.5;
```

packages/variant-theme/src/ThemeGenerator.ts
```diff
@@ -0,0 +1,426 @@
...
+  black,
+  white,
+  grey,
+  whiteAlpha,
+  blackAlpha,
+  AlphaColors,
+  type Greys,
+} from '@fluentui/react-migration-v8-v9';
+import * as lodash from 'lodash';
+import { contrast, hex_to_sRGB } from './ThemeDesigner/src/colors';
```
Hotell, line 21: based on previous feedback this should look like this:

```suggestion
import { contrast, hex_to_sRGB } from './generated-api;
```

packages/variant-theme/src/ThemeGenerator.ts
```diff
@@ -0,0 +1,426 @@
+/* eslint-disable */
+/* tslint-disable */
+//@ts-nocheck
```
Hotell, line 3:
```suggestion
```

Please don't do this - it completely defeats the purpose using typescript and eslint - if this was enabled some of my comments wouldn't be necessary because it would be caught by automation/tooling.

packages/variant-theme/package.json
Hotell, line 8: these thw packages need to be installed within monorepo root, otherwise things will not work ( we use single version policy - https://opensource.google/documentation/reference/thirdparty/oneversion )

`yarn add -DW @fluentui/scheme-utilities @fluentui/react-migration-v8-v9`

packages/variant-theme/src/ThemeGenerator.ts
```diff
@@ -0,0 +1,426 @@
...
+import {
+  getColorFromString,
+  IColor,
+  updateA
+} from '@fluentui/react';
```
Hotell, line 8: we cannot use transitive dependencies 

- update peerDeps for this project to accomodate `@fluentui/react`
- install react to monorepo `yarn add -DW @fluentui/react`

packages/variant-theme/src/ThemeGenerator.ts
```diff
@@ -0,0 +1,426 @@
...
+import {
+  black,
+  white,
+  grey,
+  whiteAlpha,
+  blackAlpha,
+  AlphaColors,
+  type Greys,
+} from '@fluentui/react-migration-v8-v9';
+import * as lodash from 'lodash';
```
Hotell, line 20: remove as explained in previous comments
```suggestion
```

packages/variant-theme/src/ThemeGenerator.ts
```diff
@@ -0,0 +1,426 @@
...
+import { contrast, hex_to_sRGB } from './ThemeDesigner/src/colors';
+
+const standardContrastRatio: number = 4.5;
+
+const greyReverse: Record<string, Greys> = lodash.invert(grey);
```
Hotell, line 25: no lodash needed

```suggestion
const greyReverse = Object.entries(grey).reduce<Record<string, Greys>>(
  (acc, [key, value]) => {
    acc[value] = key as unknown as Greys;
    return acc;
  },
  {}
);
```

packages/variant-theme/src/ThemeGenerator.ts
```diff
@@ -0,0 +1,426 @@
...
+    hex_to_sRGB(pageBackground),
+    hex_to_sRGB(color)
+  );
+  if (colorContrastWithRed < standardContrastRatio) {
+    return white;
+  }
+  return color;
+}
+
+export const calculateSoftBackground = (brandVariants, isInverted: boolean) => {
```
Hotell, line 422: export only things that are meant to be exported

```suggestion
const calculateSoftBackground = (brandVariants, isInverted: boolean) => {
```

packages/variant-theme/src/index.ts
```diff
@@ -1 +1 @@
-export {};
+export { getVariantTheme } from './ThemeGenerator';
```
Hotell, line 1: this entails non trivial amount of logic, I'd suggest adding tests for this otherwise how can you guarantee that his will work ?

Hotell (commented):
added some amount of comments

Tip for future: more focused PRs === faster merge to main ( it would be great to decouple copied code additions from fluent core and then having PR for custom logic )

With that said, let's keep this PR as is and lets start focusing on addressing comments.

ty!

packages/variant-theme/src/index.ts
mamooty: Tests are planned! Do you think they need to be in the first merge or can they be included in a follow-up?

packages/variant-theme/package.json
wenyaoj2: I have a question about the dependency rule: Which dependencies should be direct dependencies, and which should be peer dependencies? From my understanding, packages used directly should be direct dependencies. Since theme generator is using something from @fluentui/react-components, I set it as a direct dependency.

packages/variant-theme/package.json
wenyaoj2: I used VariantThemeType from scheme-utilities and some color records from react-migration-v8-v9

packages/variant-theme/src/index.ts
Hotell: it's up to the owners 🙌, I personally cannot verify if the logic works as intended so I'd require them to be added, but feel free to continue without for now.

packages/variant-theme/package.json
Hotell: right so question regarding v8-v9. I thought this package is intent is to create v8 feature parity to v9 ecosystem. is that the case or I am missing something ? ty

packages/variant-theme/package.json
Hotell: so for our case deps should look like this:

```json
{
"dependencies": {
    "@swc/helpers": "~0.5.2",
    "@fluentui/scheme-utilities": "^8.3.57",
    "@fluentui/react-migration-v8-v9": "^9.6.22",
    "@fluentui/react": "^8.119.3"
  },
  "peerDependencies": {
    "@fluentui/react-components": ">=9.46.3 <10.0.0",
    "@types/react": ">=16.8.0 <19.0.0",
    "@types/react-dom": ">=16.8.0 <19.0.0",
    "react": ">=16.8.0 <19.0.0",
    "react-dom": ">=16.8.0 <19.0.0"
  }
}
```

- @fluentui/react-migration-v8-v9 and @fluentui/react version range will always need to be aligned in this pacakge in order to not create dupes
- whilst @fluentui/react-components ships as dependency of react-migration-v8-v9 ( which is a bug ), lets follow our patterns for v9 extensions and keep it as peerDependency

Hotell (commented):
review #2 - answered question, and added some missing context. ty

packages/variant-theme/package.json
wenyaoj2: have changed package.json as your comment

packages/variant-theme/package.json
wenyaoj2: removed lodash

packages/variant-theme/package.json
wenyaoj2: I would like to use @fluentui/tokens but I found that colors are not exported from @fluentui/tokens. There is some copied files of colors in v8-v9 packages so I import these colors from v8-v9 package

packages/variant-theme/src/ThemeDesigner/src/colors/csswg.ts
wenyaoj2: Have refactored the ThemeDesigner folder

packages/variant-theme/src/ThemeGenerator.ts
wenyaoj2: I also use the VariantThemeType.None and something like it in the switch case. So I don't import it as type

packages/variant-theme/src/ThemeGenerator.ts
wenyaoj2: fixed

packages/variant-theme/src/ThemeGenerator.ts
wenyaoj2: fixed

packages/variant-theme/src/ThemeGenerator.ts
wenyaoj2: fixed

packages/variant-theme/package.json
wenyaoj2: fixed

packages/variant-theme/src/ThemeGenerator.ts
wenyaoj2: fixed

packages/variant-theme/src/ThemeGenerator.ts
wenyaoj2: removed

packages/variant-theme/src/ThemeGenerator.ts
wenyaoj2: fixed

packages/variant-theme/src/ThemeDesigner/generated-api.ts
```diff
@@ -0,0 +1,5 @@
+export * from './utils/getOverridableTokenBrandColors';
+export * from './utils/brandColors';
+export * from './utils/getBrandTokensFromPalette';
```
Hotell, line 3: we don't use these, let's re-export only api we use as we move forward with implementation in order to track what is needed from the copied

```suggestion
```

packages/variant-theme/package.json
Hotell: hmm this doesn't feel right. can you please open issue on https://github.com/microsoft/fluentui/issues so we  can avoid this in future ?

Hotell (approved):
LGTM - before merging please see my last 2 comments 
- 1 is required to address ( yarn.lock )
- 2nd is nice to have

ty

PR #206, opened 2024-07-25
ci: simplify pipelines and setup jest to follow fluent core setup

Hotell:
- removes redundant `docs` job from PR pipeline
- migratest to latest https://github.com/nrwl/nx-set-shas
- uses env variables to set NX common CLI args
- uses same jest setup as in core https://github.com/microsoft/fluentui/pull/32096 ( although there seems to be no perf gains because GHA runs on weak machine )

### DIFF
--- a/jest.preset.js
+++ b/jest.preset.js
@@ -7,4 +7,13 @@ module.exports = {
     // @see https://github.com/nrwl/nx/blob/master/packages/jest/preset/jest-preset.ts#L8-L13
     '^.+\\.(ts|js|html)$': ['@swc/jest'],
   },
+  maxWorkers: isCI() ? 1 : '50%',
 };
+
+function isCI() {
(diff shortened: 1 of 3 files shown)


jest.preset.js
```diff
@@ -7,4 +7,13 @@ module.exports = {
     // @see https://github.com/nrwl/nx/blob/master/packages/jest/preset/jest-preset.ts#L8-L13
     '^.+\\.(ts|js|html)$': ['@swc/jest'],
   },
+  maxWorkers: isCI() ? 1 : '50%',
 };
+
+function isCI() {
```
Hotell, line 13: no need to install another npm package fur such an easy task

layershifter (approved):

tudorpopams (approved):

PR #212, opened 2024-08-12
BREAKING: remove "tabIndex" assigned to handle by default

fealkhou:
- remove tabIndex: 0 from resizeHandle so consumer can set it themselves.


fealkhou:
@microsoft-github-policy-service agree [company="Microsoft"]

george-cz (approved):

PR #214, opened 2024-08-14
fix(resize-handle): make ref callback stable

layershifter:
See comments on code.

### DIFF
--- a/change/@fluentui-contrib-react-resize-handle-b566eaf8-caed-4118-b5c3-dc11216a16db.json
+++ b/change/@fluentui-contrib-react-resize-handle-b566eaf8-caed-4118-b5c3-dc11216a16db.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "fix: make ref callback stable",
+  "packageName": "@fluentui-contrib/react-resize-handle",
+  "email": "olfedias@microsoft.com",
+  "dependentChangeType": "patch"
+}


packages/react-resize-handle/src/hooks/useResizeHandle.ts
```diff
@@ -48,7 +48,7 @@ export type UseResizeHandleParams = {
    * A function that will be called to get the value that will be set as the aria-valuetext attribute on the resize handle.
    * Use this for localization.
    */
-  getA11ValueText?: (value: number) => string;
+  getA11ValueText?: (value: number) => string | undefined;
```
layershifter, line 51: This actually allows `undefined`, the default impl also returns `undefined`.

packages/react-resize-handle/src/hooks/useResizeHandle.ts
```diff
@@ -78,6 +81,11 @@ export const useResizeHandle = (params: UseResizeHandleParams) => {
   const elementRef = React.useRef<HTMLElement | null>(null);
 
   const currentValue = React.useRef(UNMEASURED);
+  const handleChange: UseResizeHandleParams['onChange'] = useEventCallback(
+    (value) => {
+      onChange?.(value);
+    }
+  );
```
layershifter, line 88: We need to use `useEventCallback()` callback there, otherwise we make memoization an implicit contract

packages/react-resize-handle/src/hooks/useResizeHandle.ts
```diff
@@ -68,8 +72,7 @@ export const useResizeHandle = (params: UseResizeHandleParams) => {
     onChange,
     onDragStart,
     onDragEnd,
-    getA11ValueText = (value) =>
-      value === UNMEASURED ? undefined : `${value.toFixed(0)}px`,
+    getA11ValueText = DEFAULT_GET_A11_VALUE_TEXT,
```
layershifter, line 75: `getA11ValueText` before was unstable reference

george-cz (approved):

PR #217, opened 2024-08-19
feat(react-interactive-tab): InteractiveTab component initial implementation

dmytrokirpa:
### Context
Previously, TabList was unable to support interactive tabs, such as tabs with a dismiss/close button to remove them, or tabs that included a dropdown or search box.

The primary slot for a Tab was a button designated with role=tab. Per ARIA guidelines, elements with role=tab are expected to only contain presentational children.

### New Behavior
This PR introduces the InteractiveTab component, which is an extension of `Tab`, utilizing a div as its base. It features three additional slots: button (where `role=tab` is applied), `beforeContent`, and `afterContent`. This structure, reminiscent of the `Input` component, permits the insertion of interactive elements both before and after the tab button.

The `InteractiveTab` applies indicator styles to the root div and button styles to the `button` slot.

### Related Issue(s)
Blocked by https://github.com/microsoft/fluentui/pull/32125
Addresses issue [#32078](https://github.com/microsoft/fluentui/issues/32078)


packages/react-interactive-tab/src/components/InteractiveTab/useTabStyles.styles.ts
dmytrokirpa, line 1: This file should be removed before merge, as we are going to import these hooks from the `@fluentui/react-tabs` once the https://github.com/microsoft/fluentui/pull/32125 merged and the package is published.

packages/react-interactive-tab/src/components/InteractiveTab/useTabAnimatedIndicator.styles.ts
dmytrokirpa, line 1: This file should be removed before merge, as we are going to import these hooks from the `@fluentui/react-tabs` once the https://github.com/microsoft/fluentui/pull/32125 merged and the package is published.

packages/react-interactive-tab/src/components/InteractiveTab/useInteractiveTabStyles.styles.ts
```diff
@@ -0,0 +1,87 @@
+import type { SlotClassNames, TabState } from '@fluentui/react-components';
+import { mergeClasses } from '@fluentui/react-components';
+
+import {
+  useTabButtonStyles_unstable,
+  useTabContentStyles_unstable,
+  useTabIndicatorStyles_unstable,
+  // Change to the `@fluentui/react-tabs` package when the new version is released
+} from './useTabStyles.styles';
```
dmytrokirpa, line 9: Change to the `@fluentui/react-tabs` package when the new version is released

packages/react-interactive-tab/stories/InteractiveTab/Default.stories.tsx
```diff
@@ -0,0 +1,66 @@
+import * as React from 'react';
```
dmytrokirpa, line 1: This file should be removed before merge, as we are going to import these hooks from the @fluentui/react-tabs once the https://github.com/microsoft/fluentui/pull/32125 merged and the package is published.

packages/react-interactive-tab/src/components/InteractiveTab/useTabAnimatedIndicator.styles.ts
```diff
@@ -0,0 +1,173 @@
+import * as React from 'react';
```
dmytrokirpa, line 1: This file should be removed before merge, as we are going to import these hooks from the @fluentui/react-tabs once the https://github.com/microsoft/fluentui/pull/32125 merged and the package is published.

packages/react-interactive-tab/package.json
```diff
@@ -2,13 +2,15 @@
   "name": "@fluentui-contrib/react-interactive-tab",
   "version": "0.0.1",
   "dependencies": {
+    "@fluentui/react-jsx-runtime": ">=9.0.29 < 10.0.0",
+    "@fluentui/react-tabs": ">=9.5.0 < 10.0.0",
```
Hotell, line 6: hmm wondering if we shouldn't rather put this to peerDeps or remove completely. the peerDep requirement will force users to install tabs under the hood anyways. if we keep it as is it might introduce react tab dupes ( impacting bundle size / potential runtime errors )

also for dependencies we don't use ranges like for peerDeps, only carets and tilda `~`/`^`

Hotell, line 6: see reasoning in previous comment

```suggestion
```

packages/react-interactive-tab/package.json
```diff
@@ -2,13 +2,15 @@
   "name": "@fluentui-contrib/react-interactive-tab",
   "version": "0.0.1",
   "dependencies": {
+    "@fluentui/react-jsx-runtime": ">=9.0.29 < 10.0.0",
```
Hotell, line 5: this one is not needed ?

```suggestion
```

packages/react-interactive-tab/src/components/InteractiveTab/InteractiveTab.test.tsx
```diff
@@ -0,0 +1,122 @@
...
+          value="1"
+          contentBefore="Before"
+          contentAfter="After"
+        >
+          Default Tab
+        </InteractiveTab>
+      </TabListProvider>
+    );
+
+    expect(result.container).toMatchSnapshot();
```
Hotell, line 99: I'm not sure what's the value of dumping these huge html structures in terms of testing quality and guaranteing that things work.

we should probably either use browser test ( via playwright ) or narrow down only to particular parts you wanna assert on.

packages/react-interactive-tab/src/components/InteractiveTab/renderInteractiveTab.tsx
```diff
@@ -0,0 +1,30 @@
+/** @jsxRuntime classic */
+/** @jsx createElement */
+
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+import { createElement } from '@fluentui/react-jsx-runtime';
```
Hotell, line 5: bit confused here, why is all of this needed ? if we wanna use custom createElement factory that should be setup either within swc or if it's only per file basis via @jsx pragma no ?

packages/react-interactive-tab/src/components/InteractiveTab/InteractiveTab.test.tsx
dmytrokirpa: I wanted to use the same tests as for the core library's Tab component to ensure consistency between the extension and the core component.

Hotell (commented):
left some comments. ty

packages/react-interactive-tab/package.json
dmytrokirpa: > note: for dependencies we don't use ranges like for peerDeps, only carets and tilda ~/^

As mentioned in another comment, I took it from the `react-chat` package from this repo.

packages/react-interactive-tab/src/components/InteractiveTab/renderInteractiveTab.tsx
dmytrokirpa: In the repository, two approaches are used: configuring the `jsx` pragma in the SWC configuration and utilizing comments. Initially, I used the latter, but have since switched to using the SWC configuration.

packages/react-interactive-tab/package.json
dmytrokirpa: There will be an eslint error regarding the imported package is not in package.json https://github.com/microsoft/fluentui-contrib/pull/217/files#diff-e57b00fdd8526b00b7a61fa87c9a06e82c8a87049bf47872dc533491ed54b2b6R3

Hotell: yes it does, but that's wrong ( gte lte makes sense only in peerDep specificiation )

the side effect of people copy pasting manually from generated peerDeps instead of automation generating these things. cc @ling1726

Hotell: yeah, in this case that's expected and we can ignore (for benefit of users) - that rule is naive in this sense as it doesn't understand dependencies

packages/react-interactive-tab/src/components/InteractiveTab/renderInteractiveTab.tsx
Hotell: yeah copy pasting "issue"

folks can you provide  guidance on custom jsx-pragma setup in contrib ? @bsunderhus @ling1726  ty

packages/react-interactive-tab/src/components/InteractiveTab/InteractiveTab.test.tsx
Hotell: yeah, so what happens in core is not the best approach on how to write test per say, but there were some historicall reason and another things that didn't make it "to stable/official guidance" 

when we have opportunity to not follow set antipatterns lets try to do that

packages/react-interactive-tab/package.json
```diff
@@ -2,13 +2,14 @@
...
```
Hotell, line 5: based on convo in previous thread

```suggestion
    "@fluentui/react-jsx-runtime": "^9.0.29",
```

packages/react-interactive-tab/stories/InteractiveTab/index.stories.tsx
```diff
@@ -0,0 +1,9 @@
+import { Meta } from '@storybook/react';
+import { InteractiveTab } from '@fluentui-contrib/react-interactive-tab';
+export { Default } from './Default.stories';
+
+const meta: Meta<typeof InteractiveTab> = {
```
Hotell, line 5: I know, copy paste from other projects in this repo. It would be great to start following community standards ( which you already started in core - which is great ! )

```
const meta = {
 ....
} satisfies Meta<typeof InteractiveTab>
```

Hotell (commented):
let's wait on folks guidance with the jsx custom pragma setup

packages/react-interactive-tab/src/components/InteractiveTab/renderInteractiveTab.tsx
bsunderhus: I'd recommend using comments @dmytrokirpa! It is the approach we're currently using on all v9 components on fluent library itself, it also makes it clear for every dev that wants to dive into our codebase that we're using something different than just the regular pragma.

> bit confused here, why is all of this needed ? if we wanna use custom createElement factory that should be setup either within swc or if it's only per file basis via @jsx pragma no ? @Hotell 

In fact, we're not event supposed to support SWC only configuration, as this will fail on our storybooks (I'm not 100% sure.... it used to fail at least the last time I checked).  We don't even want to use `createElement` in this process, we want to use the most modern `automatic` pragma where you point directly to the package that has the new runtime, but again some bundler configuration is biting us right now.

Hotell: https://github.com/microsoft/fluentui/pull/32838/files

dmytrokirpa: Reverted back to use comment pragma

packages/react-interactive-tab/package.json
dmytrokirpa: The `check-dependencies` task fails on CI, we might need to update other packages to this format in one go. I'd prefer to do this in scope of other PR, not related to the specific pacakage.

image

packages/react-interactive-tab/src/components/InteractiveTab/renderInteractiveTab.tsx
```diff
@@ -0,0 +1,30 @@
+/** @jsx createElement */
```
Hotell, line 1: lets use react v17 features ( we are using v17 ) - aligns also with core 

https://github.com/microsoft/fluentui/pull/32838/files#diff-07b5fa9a3e830a5d704eea58256b26d63c6af7536119dbd16c969536ef3505a9R25

Hotell (approved):
1 actionable - for the jsx pragma

ty

packages/react-interactive-tab/package.json
Hotell: yeah that sounds good, in order to pass pipeline you'll need to ignore your package deps from syncpack

PR #221, opened 2024-08-19
feat(react-keytips): initial implementation 

mainframev:
- [Design spec in Figma](https://www.figma.com/design/EMvdJTxEQMCqQHpgHdcfua/Keytips?node-id=1799-1343&node-type=canvas&t=QvOhfccxGuPdG7mN-0).
- [Component spec](https://github.com/microsoft/fluentui-contrib/blob/main/packages/react-keytips/src/docs/Spec.md).
- [Keytips in v8](https://developer.microsoft.com/en-us/fluentui#/controls/web/keytips).

Support for persistent keytips will be added later. This PR covers draft and initial implementation.


packages/react-keytips/src/components/Keytips/useKeytips.tsx
```diff
@@ -0,0 +1,261 @@
...
+    };
+  }, [inKeytipMode, handleExitKeytipMode]);
+
+  const visibleKeytips = Object.entries(keytips)
+    .filter(([, { visible, visibleInternal }]) => visible || visibleInternal)
+    .map(([keytipId, keytipProps]) => (
+      <Keytip key={keytipId} {...keytipProps} />
+    ));
+
+  const hiddenKeytips = Object.values(keytips).map(({ keySequences }) => (
```
mainframev, line 236: this is similar to v8 keytips implementation and needed for maintaining ids for `aria-describedby`

packages/react-keytips/src/components/Keytips/useKeytips.tsx
```diff
@@ -0,0 +1,274 @@
...
+      targetDocument?.defaultView?.removeEventListener('resize', handleDismiss);
+      targetDocument?.defaultView?.removeEventListener('scroll', handleDismiss);
+    };
+
+    addEventListeners();
+    return removeEventListeners;
+  }, [targetDocument, inKeytipMode, handleExitKeytipMode]);
+
+  React.useEffect(() => {
+    const handleKeyDown = (ev: KeyboardEvent) => {
```
dmytrokirpa, line 194: The complexity of this method is quite high; it might be beneficial to break it down into smaller, named functions.

packages/react-keytips/package.json
```diff
@@ -2,7 +2,11 @@
   "name": "@fluentui-contrib/react-keytips",
   "version": "0.0.1",
   "dependencies": {
-    "@swc/helpers": "~0.5.2"
+    "@swc/helpers": "~0.5.2",
+    "@fluentui/react-jsx-runtime": ">=9.0.29 < 10.0.0",
+    "@fluentui/react-utilities": ">=9.16.0 < 10.0.0",
+    "@fluentui/react-positioning": ">=9.16.0 < 10.0.0",
```
dmytrokirpa, line 8: We shouldn't use ranges in prod dependencies
```suggestion
    "@fluentui/react-positioning": "^9.16.0",
```

packages/react-keytips/src/components/Keytip/Keytip.types.ts
```diff
@@ -0,0 +1,81 @@
...
+>;
+
+export type ReturnKeytipEventHandler<E = HTMLElement> = EventHandler<
+  EventData<'keydown', KeyboardEvent> & {
+    targetElement: E;
+  }
+>;
+
+export type OnVisibleChangeData = {
+  visible: boolean;
```
dmytrokirpa, line 32: Please add prop description

packages/react-keytips/src/components/Keytips/Keytips.test.tsx
```diff
@@ -0,0 +1,12 @@
+import * as React from 'react';
+import { render } from '@testing-library/react';
+import { Keytips } from './Keytips';
+
+describe('Keytips', () => {
+  // TODO add more tests here, and create visual regression tests in /apps/vr-tests
+
+  it('renders a default state', () => {
+    const result = render(<Keytips content="Alt Windows" />);
+    expect(result.container).toMatchSnapshot();
```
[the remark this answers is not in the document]
mainframev: Yep, will do, this test was generated by scaffolding. I though to add tests separately from initial PR, according to our guideline for component implementation. ;)

packages/react-keytips/package.json
mainframev: thx. will change it for react-positioning as u suggested, other deps with ranges are has to be fixed in other packages as well, otherwise CI will fail on syncpack check

packages/react-keytips/src/hooks/useKeytipRef.ts
```diff
@@ -0,0 +1,54 @@
...
+    positioning: {
+      target: node,
+      ...keytip.positioning,
+    },
+  };
+
+  const prevKeytip = usePrevious(ktp);
+
+  const setRef = React.useCallback(
+    (node?: T) => {
```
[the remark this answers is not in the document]
mainframev: Good catch, left from pervious iterations where it was different

packages/react-keytips/src/hooks/useTree.ts
```diff
@@ -0,0 +1,168 @@
...
+    []
+  );
+
+  const nodeMap = React.useRef<Map<string, KeytipTreeNode>>(
+    new Map([[KTP_ROOT_ID, root]])
+  );
+
+  const currentKeytip = React.useRef<KeytipTreeNode | undefined>();
+
+  const addNode = React.useCallback((keytip: KeytipWithId) => {
```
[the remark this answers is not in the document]
mainframev: @dmytrokirpa Thanks for suggestion. Tbh, not sure, if it will make it easier to test, it will be similar to use useState so I will need to wrap all these methods in tests with `act`, the current implementation with `useRef` also avoids re-renders and not sure if having state here is really needed 🙂

packages/react-keytips/src/components/Keytip/renderKeytip.tsx
```diff
@@ -0,0 +1,16 @@
+/** @jsxRuntime classic */
```
dmytrokirpa, line 1: 1. Change pragmas to:

```
/** @jsxRuntime automatic */
/** @jsxImportSource @fluentui/react-jsx-runtime */
```

2. Remove `import { createElement } from '@fluentui/react-jsx-runtime';`

More details https://react.fluentui.dev/?path=/docs/concepts-developer-customizing-components-with-slots--docs#rendering-components-with-slots

packages/react-keytips/src/hooks/useKeytipRef.ts
```diff
@@ -0,0 +1,53 @@
...
+
+  const prevKeytip = usePrevious(ktp);
+
+  React.useEffect(() => {
+    if (prevKeytip) {
+      if (!isEqualArray(prevKeytip.keySequences, ktp.keySequences)) {
+        dispatch(EVENTS.KEYTIP_UPDATED, ktp);
+      }
+    }
+  });
```
dmytrokirpa, line 42: Effect dependencies are not specified

packages/react-keytips/src/hooks/useTree.ts
```diff
@@ -0,0 +1,174 @@
...
+
+  const updateNode = React.useCallback((keytip: KeytipWithId) => {
+    const node = nodeMap.current.get(keytip.uniqueId);
+
+    if (node) {
+      const prevParent = node.parent;
+      const parent = getParentID(keytip.keySequences);
+
+      if (prevParent !== parent) {
+        [...nodeMap.current.values()].forEach((node) => {
```
dmytrokirpa, line 84:
```suggestion
        for (const node of nodeMap.current.values()) {
```

packages/react-keytips/src/hooks/useKeytipRef.ts
mainframev: this is on purpose, so this check runs every time ;) in v8 it was kinda similar [for update logic](https://github.com/microsoft/fluentui/blob/master/packages/react/src/components/KeytipData/useKeytipData.ts#L29)

packages/react-keytips/package.json
```diff
@@ -13,5 +10,11 @@
     "@types/react-dom": ">=16.8.0 <19.0.0",
     "react": ">=16.8.0 <19.0.0",
     "react-dom": ">=16.8.0 <19.0.0"
+  },
+  "dependencies": {
+    "@fluentui/react-jsx-runtime": ">=9.0.29 < 10.0.0",
+    "@fluentui/react-utilities": ">=9.16.0 < 10.0.0",
+    "@fluentui/react-positioning": "^9.16.0",
+    "@fluentui/keyboard-keys": "~9.0.6"
```
Hotell, line 18: which of these apis that this library is using are not part of react-components suite ?

if api's are accessible via rc suite those sub-packages should not be specified as dependencies.

**what needs to change** are the versioning ranges, `gte` etc is supposed to be used only for peerDependencies, anything else needs to use `~` or `^` depending on the used package semver policy

cc @dmytrokirpa ( we talked about this inconsistency in whole contrib - not sure if you managed to create that issue yet )

packages/react-keytips/src/components/Keytip/renderKeytip.tsx
Hotell: we will have rule for this soon TM https://github.com/microsoft/fluentui/pull/32842

packages/react-keytips/src/components/Keytips/Keytips.test.tsx
```diff
@@ -0,0 +1,192 @@
+import * as React from 'react';
+import { render, screen, act } from '@testing-library/react';
+import { Keytips } from './Keytips';
+import { useKeytipRef } from '../..';
```
Hotell, line 4: I suggest being explicit by providing module from which you're importing. while this work now it wont work with ESM in future. also importing from barrel files === slower test execution

packages/react-keytips/src/components/Keytips/Keytips.test.tsx
```diff
@@ -0,0 +1,192 @@
...
+    const portal = baseElement.querySelector('[data-portal-node]');
+
+    expect(portal?.children).toHaveLength(3);
+
+    expect(screen.queryAllByRole('tooltip')).toHaveLength(0);
+
+    // enter keytip mode
+    await act(async () => await userEvent.keyboard('{Alt>}{Meta}'));
+
+    expect(screen.getAllByRole('tooltip')).toHaveLength(2);
```
Hotell, line 52: this whole package is crucial to work correctly in real browsers (focus/keypresses etc) -> I'd suggest to add playwright component test and cover the real functionality there

packages/react-keytips/src/hooks/useTree.ts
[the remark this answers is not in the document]
mainframev: yep, but I dont want it to be stateful :)

packages/react-keytips/src/components/Keytip/renderKeytip.tsx
mainframev: discussed offline, context: [this](https://github.com/microsoft/fluentui-contrib/pull/217#discussion_r1755018804) and current storybook webpack & babel aren't working with automatic pragma, needs investigation.

packages/react-keytips/src/hooks/useKeytipRef.ts
dmytrokirpa: Could you please add a note stating it's done on purpose? This way, we can avoid similar questions in future 😄

packages/react-keytips/src/hooks/useTree.ts
```diff
@@ -0,0 +1,175 @@
...
+  }, []);
+
+  const getMatchingNode = React.useCallback(
+    (sequence: string) => {
+      const { current } = currentKeytip;
+      if (!current) {
+        return undefined;
+      }
+
+      const matchingNodes = [...current.children]
```
dmytrokirpa, line 109: Why do we need to use the spread operator here?

packages/react-keytips/stories/index.stories.ts
```diff
@@ -0,0 +1,13 @@
...
+import { Keytips } from '@fluentui-contrib/react-keytips';
+export { DefaultStory as Default } from './Default.stories';
+export { WithTabsStory as WithTabs } from './WithTabs.stories';
+export { DynamicStory as Dynamic } from './Dynamic.stories';
+export { OverflowStory as Overflow } from './OverflowMenu.stories';
+
+const meta: Meta<typeof Keytips> = {
+  title: 'Keytips',
+  component: Keytips,
+};
```
dmytrokirpa, line 11: nit: `satisfies Meta<typeof Keytips>`

dmytrokirpa (approved):

packages/react-keytips/src/hooks/useTree.ts
mainframev: because it's a Set

dmytrokirpa: I suggest opting for `Array.from(current.children)` as it enhances readability and more effectively communicates the purpose, compared to using the spread operator.

packages/react-keytips/src/components/Keytip/renderKeytip.tsx
Hotell: can you please open issue for this ? we don't wanna use classic runtime going forward

packages/react-keytips/package.json
Hotell: can you please open issue on this topic in core repo ?

change/@fluentui-contrib-react-keytips-2c77a34b-1b5c-4d63-a244-ad06854777c9.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "feat(react-keytips): initial implementation",
```
Hotell, line 3: the "scope" is covered by the change-file 


```suggestion
  "comment": "feat: initial implementation",
```

packages/react-keytips/package.json
```diff
@@ -1,9 +1,6 @@
 {
   "name": "@fluentui-contrib/react-keytips",
   "version": "0.0.1",
-  "dependencies": {
-    "@swc/helpers": "~0.5.2"
```
Hotell, line 5: what happened here ? without this dep it will explode in user land

mainframev: Good spot, thanks, will get it back

Hotell (approved):

PR #223, opened 2024-08-19
feat(react-resize-handle): Use slider role for better screen reader support

george-cz:
After doing research about the support of `separator` and `slider` roles across different screen readers and platforms it became obvious that the slider role has much better narration and keyboard support.

While `separator` fits the use case the best on paper, in real-world it's support leaves much to be desired.

To summarize the issues with `separator`:
 - aria-label not supported on VoiceOver on Mac for `separator`
 - VO keys not working for `separator`
 - current value (aria-valuetext) not being narrated in Jaws
 - no mode auto switching in NVDA and Jaws, hindering the experience.

All of these issues are avoided by using `slider`.

"Horizontal" and "vertical" orientation needs to be flipped for "slider".

### DIFF
--- a/packages/react-resize-handle/src/hooks/useResizeHandle.component-browser-spec.tsx
+++ b/packages/react-resize-handle/src/hooks/useResizeHandle.component-browser-spec.tsx
@@ -5,21 +5,78 @@ import { TestArea } from './useResizeHandleExample.component-browser-spec';
 
 test.use({ viewport: { width: 500, height: 500 } });
 
+async function mountTest(mount) {
+  const component = await mount(<TestArea />);
+  const dragHandle = component.getByRole('slider');
+  const wrapper = component.getByTestId('wrapper');
+  const valueEl = component.getByTestId('value');
+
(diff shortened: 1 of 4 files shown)


packages/react-resize-handle/src/hooks/useResizeHandle.component-browser-spec.tsx
```diff
@@ -5,21 +5,78 @@ import { TestArea } from './useResizeHandleExample.component-browser-spec';
...
+    await validateComponent(component, 50);
+  });
+
+  test.describe("min/max value can't be exceeded", () => {
+    test('with mouse', async ({ mount, page }) => {
+      const component = await mountTest(mount);
+      await component.dragHandle.hover();
+      await page.mouse.down();
+      await page.mouse.move(1000, 0);
+      await page.mouse.up();
```
layershifter, line 54: nit: create a helper for drag?

layershifter (approved):

PR #226, opened 2024-08-22
Fix: pass event to onValueChange

george-cz:
This fixes an issue where the consumer wouldn't know if the `onValueChange` was triggered by a mouse or by keyboard. They might decide to react on the user finishing resizing, which, for a mouse is easy (`onDragEnd`) but for a keyboard we don't have such callback, since we trigger `onValueChange` each time a button is pressed and value updated.

By updating the event callbacks to be in better alignment with Fluent UI and using `EventHandler`, we can now determine if the source of `onValueChange` is keyboard by comparing the `data.type` attribute to `EVENTS.keyboard`


packages/react-resize-handle/src/hooks/useResizeHandle.ts
```diff
@@ -84,37 +87,42 @@ export const useResizeHandle = (params: UseResizeHandleParams) => {
...
-    // Make sure to only apply the value if it's not the initial value!
-    if (currentValue.current !== UNMEASURED) {
-      wrapperRef.current?.style.setProperty(
-        variableName,
-        `${currentValue.current}px`
-      );
-      onChange?.(currentValue.current);
-    }
-  }, [getA11ValueText, maxValue, minValue, onChange, variableName]);
+  const updateElementsAttrs = React.useCallback(
```
george-cz, line 90: pls hide whitespace changes, as the formatting was a bit wild for some reason

packages/react-resize-handle/src/hooks/useKeyboardHandler.ts
```diff
@@ -3,7 +3,7 @@ import * as React from 'react';
 import { GrowDirection, SupportedKeys } from '../types';
 
 export type UseKeyboardHandlerOptions = {
-  onValueChange: (value: number) => void;
+  onValueChange: (value: number, event: KeyboardEvent) => void;
```
layershifter, line 6: Let's rather break it and use `EventHandler` type like in Fluent to avoid confusion. It also will be more explicit about interactions.

packages/react-resize-handle/stories/Default.stories.tsx
```diff
@@ -75,7 +75,8 @@ const Component = (props: ComponentProps) => {
     variableName: NAV_SIZE_CSS_VAR,
     growDirection: 'end',
     relative: true,
-    onChange: (value: number) => {
+    onChange: (value: number, event) => {
+      console.log(event);
```
layershifter, line 79: Oops

packages/react-resize-handle/src/hooks/useMouseHandler.ts
```diff
@@ -12,7 +12,7 @@ export type UseMouseHandlerParams = {
   onDown?: (event: NativeTouchOrMouseEvent) => void;
   onMove?: (event: NativeTouchOrMouseEvent) => void;
   growDirection: GrowDirection;
-  onValueChange: (value: number) => void;
+  onValueChange: (value: number, event: NativeTouchOrMouseEvent) => void;
```
layershifter, line 15: Let's use `EventHandler` type there, too

layershifter (changes requested):

packages/react-resize-handle/src/hooks/useKeyboardHandler.ts
george-cz: I started working on exactly this before the weekend :)

packages/react-resize-handle/stories/Default.stories.tsx
```diff
@@ -75,13 +75,13 @@ const Component = (props: ComponentProps) => {
     variableName: NAV_SIZE_CSS_VAR,
     growDirection: 'end',
     relative: true,
-    onChange: (value: number) => {
-      props.onChange(value);
+    onChange: (_, data) => {
+      console.log(data);
```
layershifter, line 79: Is it intentional?

packages/react-resize-handle/src/hooks/useKeyboardHandler.ts
layershifter, line 6: Let's flip arguments even there

packages/react-resize-handle/src/hooks/useResizeHandle.ts
```diff
@@ -61,6 +76,38 @@ export type UseResizeHandleParams = {
   relative?: boolean;
 };
 
+const getUpdateEventData = (
+  event: NativeTouchOrMouseEvent | KeyboardEvent | CustomEvent,
+  value: number
+): ResizeHandleUpdateEventData => {
+  switch (event.constructor) {
+    case TouchEvent:
```
layershifter, line 84: We can get a problem as there similar to this one:

https://github.com/microsoft/fluentui/blob/5c671b32c10347f647b09105695f840824ebee4e/packages/react-components/react-utilities/src/utils/isHTMLElement.ts#L2-L6

I.e. could be broken for multi `window` scenarios as `TouchEvent` is actualy `window.TouchEvent`

packages/react-resize-handle/src/hooks/useResizeHandle.ts
```diff
@@ -15,6 +15,22 @@ const clampWithMode = (
   return relative || value === UNMEASURED ? value : clamp(value, min, max);
 };
 
+import { EventData, EventHandler } from '@fluentui/react-utilities';
```
layershifter, line 18:
```suggestion
import type { EventData, EventHandler } from '@fluentui/react-utilities';
```

packages/react-resize-handle/src/hooks/useResizeHandle.ts
```diff
@@ -84,37 +131,47 @@ export const useResizeHandle = (params: UseResizeHandleParams) => {
...
+      if (currentValue.current !== UNMEASURED) {
+        wrapperRef.current?.style.setProperty(
+          variableName,
+          `${currentValue.current}px`
+        );
+
+        const eventToPass = event || new CustomEvent(EVENTS.setValue);
+        onChange?.(
+          eventToPass,
+          getUpdateEventData(eventToPass, currentValue.current)
```
layershifter, line 169:
```suggestion
        onChange?.(
          eventToPass,
          getUpdateEventData(event, currentValue.current)
```

Should `getUpdateEventData()` create event instead?

packages/react-resize-handle/src/hooks/useResizeHandle.ts
george-cz: okay I have an idea how to fix this without the construtor

packages/react-resize-handle/src/hooks/useResizeHandle.ts
george-cz: that won't work, as `onChange` is an `EventHandler` and it expects the first parameter to be of type `Event`, which is also not nullable

layershifter: Did not get it. Why it cannot be following?

```ts
const getUpdateEventData = (
  event: NativeTouchOrMouseEvent | KeyboardEvent | null,
  value: number
): ResizeHandleUpdateEventData => {
  // ...
    default:
      return {
        type: EVENTS.setValue,
        value,
        event: new CustomEvent(EVENTS.setValue),
      };
}
```

packages/react-resize-handle/src/types.ts
```diff
@@ -1,3 +1,5 @@
+import { EventData } from '@fluentui/react-utilities';
```
layershifter, line 1:
```suggestion
import type { EventData } from '@fluentui/react-utilities';
```

packages/react-resize-handle/src/hooks/useMouseHandler.ts
```diff
@@ -98,7 +122,18 @@ export function useMouseHandler(params: UseMouseHandlerParams) {
...
+      'touches' in event
+        ? {
+            type: EVENTS.touch,
+            event: event as TouchEvent,
+          }
+        : {
+            type: EVENTS.mouse,
+            event: event as MouseEvent,
+          }
+    );
```
layershifter, line 136: nit: probably we can add a specific local helper for that condition ('touches' in event)

PR #228, opened 2024-08-29
docs(react-keytips): add specification

mainframev:
Initial specification for `Keytips`, might be adjusted after further implementation changes. 
[Design spec in Figma ](https://www.figma.com/design/EMvdJTxEQMCqQHpgHdcfua/Keytips?node-id=7852-5113&t=0iyLoie8uLNbCDN3-1)


packages/react-keytips/src/docs/Spec.md
```diff
@@ -0,0 +1,191 @@
+# @fluentui-contrib/react-keytips Spec
+
+## Background
+
+A `Keytip` is a small Tooltip near a component that indicates a key sequence that will trigger that component. There are
+not to be confused with keyboard shortcuts, they are instead key sequence to traverse through levels of UI components.
+
+## Prior Art
+
+- [Convergence Epic](https://github.com/microsoft/fluentui/issues/26420)
```
ValentinaKozlova, line 10: This link leads to Drawer component. I don't see Epic for Keytips. Not sure if we need it for contrib components

packages/react-keytips/src/docs/Spec.md
```diff
@@ -0,0 +1,191 @@
...
+It provides `KeytipLayer` and `useKeytipRef` to manage the keytips.
+
+- [KeytipLayer](https://developer.microsoft.com/en-us/fluentui#/controls/web/keytips#IKeytipLayer) - Helps to register a
+  key sequence for entering and exiting from keytips, for going back to the previous level of keytips, provides necessary callbacks.
+- [useKeytipRef](https://github.com/microsoft/fluentui/blob/master/packages/react/src/components/KeytipData/useKeytipRef.ts) - Creates a ref,
+  which is used for passing to `Keytip` target element, handles setting necessary attributes.
+
+```tsx
+export default function App() {
+  // default is 'Alt-Windows' and 'Option-Control' on macOS
```
ValentinaKozlova, line 24:
```suggestion
  // default is 'Alt-Windows' on Windows and 'Option-Control' on macOS
```

packages/react-keytips/src/docs/Spec.md
mainframev: thanks! I will keep it in case if somebody will later would like to look up for epic. ;)

packages/react-keytips/src/docs/Spec.md
```diff
@@ -0,0 +1,190 @@
...
+    </>
+};
+```
+
+## Variants
+
+### Positioning Variants
+
+The `Keytips` component using [positioning API](https://react.fluentui.dev/?path=/docs/concepts-developer-positioning-components--docs) and can be controlled by `positioning` prop in `useKeytipRef` hook.
+The keytip is positioned below and centered to the target element by default.
```
ValentinaKozlova, line 101: Maybe it'll be good to include other positioning variants too, not only default one

packages/react-keytips/src/docs/Spec.md
```diff
@@ -0,0 +1,190 @@
...
+
+![Keytip positioning example](assets/positioning.png)
+
+### Style variants
+
+`Keytips` have two style variants (appearance): `normal` and `inverted` (default).
```
ValentinaKozlova, line 107: Where is appearance in the design? I see there's only dark and light theme. But it should change the colors automatically

packages/react-keytips/src/docs/Spec.md
```diff
@@ -0,0 +1,190 @@
...
+## API
+
+### Keytips
+
+| Prop Name            | Type                                   | Default         | Description                                                     |
+| -------------------- | -------------------------------------- | --------------- | --------------------------------------------------------------- |
+| `mountNode`          |                                        |                 | Where the Portal children are mounted on DOM.                   |
+| `startSequence`      | `string`                               | `'alt+control'` | Key sequence that will start keytips mode.                      |
+| `returnSequence`     | `string`                               | `'escape'`      | Key sequences that execute the return functionality in keytips. |
+| `exitSequence`       | `string`                               |                 | Key sequences that will exit keytips mode.                      |
```
ValentinaKozlova, line 120: Here's also `escape` default?

packages/react-keytips/src/docs/Spec.md
```diff
@@ -0,0 +1,190 @@
...
+| `visible`      | `boolean`                   | `false`                                  | Control the Keytip's visibility programmatically.                                                                                                                 |
+| `content`      | `string`                    |                                          | The text content of the Keytip.                                                                                                                                   |
+
+## Structure
+
+### JSX Tree
+
+```tsx
+<Portal>
+  <Keytip content="A" positioning={{ target }} />
```
ValentinaKozlova, line 143: If targets are different maybe to mention something like `target1`, `target2` and `target3`

packages/react-keytips/src/docs/Spec.md
mainframev: Design spec does not mention explicitly if it should be controlled or not, similar in Tooltip, there are two variants, but Tooltip provides `appearance` prop. I will probably need to ask designer, or WDYT? Probably should be only automatically.

packages/react-keytips/src/docs/Spec.md
```diff
@@ -0,0 +1,190 @@
...
+
+## Migration
+
+See [MIGRATION.md](./MIGRATION.md).
+
+## Behaviors
+
+## Entering and Exiting Keytips Mode
+
+- When the `startSequence` is pressed, the keytips component will enter keytips mode. By default the `startSequence` is `'alt+control'` (option+control on macOS).
```
ValentinaKozlova, line 168:
```suggestion
- When the `startSequence` is pressed, the keytips component will enter keytips mode. By default the `startSequence` is `'alt+Windows'` (option+control on macOS).
```

packages/react-keytips/src/docs/Spec.md
mainframev: `escape` is already set for `returnSequence`, so it will exit from keytip mode only on the root level of the Keytips. So for the `exitSequence` we won't probably need a default. It will behave like this -> if keytip mode is on, then pressing `startSequence` again will exit from keytip mode and `escape` behave as I mentioned above. If there will be a need for user for specific key sequence for exit, it's possible to add it via this prop :)

packages/react-keytips/src/docs/Spec.md
```diff
@@ -0,0 +1,190 @@
...
+- Keytip sequence is case-insensitive and limited to 3 characters.
+
+## Keytip on disabled item
+
+- Unlike Fabric, the Keytip won't appear for disabled target.
+
+## Keytip positioning
+
+Keytip component using [positioning API](https://react.fluentui.dev/?path=/docs/concepts-developer-positioning-components--docs) and
+can be controlled by `positioning` prop in `useKeytipRef` hook. By default the keytip is positioned below and centered to the target element.
```
ValentinaKozlova, line 190: What about accessibility section? What screen reader should announce for keytips? And what `area-attributes` should be here?

packages/react-keytips/src/docs/Spec.md
mainframev: @ValentinaKozlova I guess the suggestion is not correct, it should be alt+control / option-control, `win` key is different one.

packages/react-keytips/src/docs/Spec.md
ValentinaKozlova: I think it's better to ask the designer. Maybe they forgot to include it. For example for Breadcrumb we had a link on Button component but in the Breadcrumb's spec the design was slightly different.

packages/react-keytips/src/docs/Spec.md
```diff
@@ -0,0 +1,190 @@
...
+![Keytip style variants example](assets/style-variants.png)
+
+## API
+
+### Keytips
+
+| Prop Name            | Type                                   | Default         | Description                                                     |
+| -------------------- | -------------------------------------- | --------------- | --------------------------------------------------------------- |
+| `mountNode`          |                                        |                 | Where the Portal children are mounted on DOM.                   |
+| `startSequence`      | `string`                               | `'alt+control'` | Key sequence that will start keytips mode.                      |
```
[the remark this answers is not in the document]
mainframev: do not think so, at least in v8 it was set on `alt+control` / `option+control`, so I do not think we should change it

packages/react-keytips/src/docs/Spec.md
ValentinaKozlova: In the design spec it says Windows, and in V8 on Windows keytips are triggered by `alt+windows` not by `alt+control`

packages/react-keytips/src/docs/Spec.md
mainframev: Need to investigate and play around with different area attributes ask designer and Sarah. I will update it to this spec ASAP ;)

packages/react-keytips/src/docs/Spec.md
mainframev: Already asked ;) Thanks!

packages/react-keytips/src/docs/Spec.md
mainframev: right, good catch then :) Updated

ValentinaKozlova (approved):

packages/react-keytips/src/docs/Spec.md
```diff
@@ -0,0 +1,203 @@
...
+
+In terms of accessibility, `Keytip` component is very similar to `Tooltip` - it's a small popup that displays information related to an element, with difference, that `Keytip` can't be displayed by mouse hover over target element. Multiple keytips can be visible at the same time.
+
+- Each `Keytip` is assigned the role `"tooltip"`.
+- The `Keytip` target element references the corresponding keytip element using the `"aria-describedby"` attribute, providing a clear association between the target and the keytip (tooltip).
+- `Keytips` component has `content` prop, which is responsible for adding the `"aria-describedby"` for default start key sequence.
+- `Keytip` is not focusable.
+- `Keytip` adds `"aria-keyshortcuts"` attribute to the target element, which contains the key sequence that will trigger the keytip.
+- Unlike Tooltip, pressing `Escape` does not always hide the `Keytip`. By default, `Keytip` is configured to use `Escape` as a return sequence. Therefore, if there are ancestor keytips, pressing it will not hide the current `Keytip` but will instead display the ancestor keytips.
+- Focus stays on the trigger element while `Keytip` is displayed.
```
mainframev, line 203: @smhigley 👋🏻 

Here are a couple of a11y things I am not sure about: 
- Can `Keytip` have role `"tooltip"`?  (in the previous version there was no role for Keytip)
- In `v8` Keytip was adding a `aria-describedby` on the target element with id of the root layer and current keytip (so it's announcing start sequence and sequence of current keytip). 
-  In the current draft implementation I added both, so after the `aria-describedby` announced it will announce`shortcuts available`  of `aria-keyshortcuts`. So, question if we can have both? or if not, which way is better in your opinion? 

I'd be very grateful if you could take a look on this :)

PR #238, opened 2024-09-23
add @stortybook/addon-themes with default themes to packages

mainframev:


package.json
```diff
@@ -52,6 +52,7 @@
     "@storybook/addon-essentials": "7.6.10",
     "@storybook/addon-interactions": "^7.5.3",
     "@storybook/addon-storysource": "7.6.10",
+    "@storybook/addon-themes": "7.6.17",
```
Hotell, line 55: all sb versions needs to be on same version - please bump other packages as well

packages/nx-plugin/src/generators/configure-storybook/files/.storybook/main.ts.template
```diff
@@ -4,6 +4,9 @@ const config: StorybookConfig = {
   stories: ['../stories/**/index.stories.@(js|jsx|ts|tsx|mdx)'],
   addons: [
     '@nx/react/plugins/storybook',
+    '@storybook/addon-docs',
```
Hotell, line 7: seems this adds more than "advertised" within the PR 👀

- why do we need actions ? is anyone using that ( from my personal experiences using those might be an anti-pattern )
- what about the docs addon ?

packages/nx-plugin/src/generators/configure-storybook/files/.storybook/preview.tsx.template
```diff
@@ -1,18 +1,29 @@
...
+} from '@fluentui/react-components';
 
 const preview: Preview = {
   decorators: [
-    (Story) => (
-      <FluentProvider theme={webLightTheme}>
-        <Story />
-      </FluentProvider>
-    ),
+    withThemeFromJSXProvider({
```
Hotell, line 15: this starts to become a maintenance burden / hard to manage boilerplate -  this should be implemented as part of sb plugin ( we  have this in core repo AFAIR ? -> it will be published and streamlined here as well in future )

I'd start just with adding this functionality to your package in order to not introduce unmaintained boilerplate while accommodating your package needs. wdyt?

packages/pierce-dom/.eslintrc.json
```diff
@@ -1,6 +1,6 @@
 {
   "extends": ["../../.eslintrc.json"],
-  "ignorePatterns": ["!**/*"],
+  "ignorePatterns": ["!**/*", "storybook-static"],
```
Hotell, line 3: why is this needed ?

packages/react-themeless-provider/tsconfig.storybook.json
Hotell, line 1: what happened here ?

Hotell (commented):
action required - ty

packages/pierce-dom/.eslintrc.json
mainframev: was added automatically when I run generator, will check why it was added

packages/nx-plugin/src/generators/configure-storybook/files/.storybook/preview.tsx.template
mainframev: I thought about this as well, but after noticed it's not published and was't sure if it's planned to be published, decided to use this approach as for contrib that might be sufficient to use official sb plugin for theming. Will close this one then. Thanks.

Hotell (approved):
please make sure to provide description of your PR. ty

PR #251, opened 2024-10-18
feat(react-keytips): add invokeEvent prop

mainframev:
Feature was requested earlier for v8: https://github.com/microsoft/fluentui/issues/31499.

### DIFF
--- a/change/@fluentui-contrib-react-keytips-6c388a16-92f8-4352-a0e6-37f7bdd0eb14.json
+++ b/change/@fluentui-contrib-react-keytips-6c388a16-92f8-4352-a0e6-37f7bdd0eb14.json
@@ -1,5 +1,5 @@
 {
-  "type": "minor",
+  "type": "patch",
   "comment": "feat: add invokeEvent prop",
   "packageName": "@fluentui-contrib/react-keytips",
   "email": "vgenaev@gmail.com",
--- a/change/@fluentui-contrib-react-keytips-6c388a16-92f8-4352-a0e6-37f7bdd0eb14.json
+++ b/change/@fluentui-contrib-react-keytips-6c388a16-92f8-4352-a0e6-37f7bdd0eb14.json
@@ -1,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "feat: add invokeEvent prop",
+  "packageName": "@fluentui-contrib/react-keytips",
+  "email": "vgenaev@gmail.com",
+  "dependentChangeType": "patch"
+}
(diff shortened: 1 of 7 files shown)


change/@fluentui-contrib-react-keytips-6c388a16-92f8-4352-a0e6-37f7bdd0eb14.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
```
Hotell, line 2: this is wrong - we operate in zero major semver policy

- major - not allowed
- minor - breaking changes
- patch - feature/fix

```suggestion
  "type": "patch",
```

packages/react-keytips/src/components/Keytips/Keytips.test.tsx
```diff
@@ -116,4 +116,40 @@ describe('Keytips', () => {
     expect(screen.queryAllByRole('tooltip')).toHaveLength(0);
     expect(onExitKeytipsMode).toHaveBeenCalledTimes(1);
   });
+
+  it('should support different invoking event', async () => {
```
dmytrokirpa, line 120: In my opinion, the current test focuses more on the implementation rather than the behavior it's supposed to assess. I recommend substituting it with two Playwright tests that specifically evaluate the functionality in both keydown and keyup modes. What do you think?

packages/react-keytips/src/hooks/useHotkeys.ts
```diff
@@ -54,7 +54,11 @@ const isKeyMatchingKeyboardEvent = (
   return false;
 };
 
-export const useHotkeys = (hotkeys: Hotkey[], target?: Document) => {
+export const useHotkeys = (
+  hotkeys: Hotkey[],
+  invokeEvent: 'keyup' | 'keydown' = 'keydown',
```
dmytrokirpa, line 59: nit: this type could potentially be reused for component properties and the hook.

packages/react-keytips/src/components/Keytips/Keytips.test.tsx
mainframev: not sure if there are two tests needed, keydown is already covered by basic testing in playwright, it would be a duplication. Although I can add the keyup test to test it in playwright, instead of testing-library, as it's now

mainframev: Added, I don't mind having both things tested, but having only keyup test cases should be enough, as keydown tested by other test cases.

packages/react-keytips/src/hooks/useHotkeys.ts
```diff
@@ -54,7 +55,11 @@ const isKeyMatchingKeyboardEvent = (
   return false;
 };
 
-export const useHotkeys = (hotkeys: Hotkey[], target?: Document) => {
+export const useHotkeys = (
+  hotkeys: Hotkey[],
+  invokeEvent: KeytipsProps['invokeEvent'] = 'keydown',
```
dmytrokirpa, line 60: nit: Consider creating a dedicated type `InvokeEvent = 'keydown' | 'keyup'` that could be used both as a parameter for a hook and as a prop type, rather than directly extracting it from the props.

dmytrokirpa (approved):

PR #254, opened 2024-10-25
docs(react-keytips): improve keytips for docsite

mainframev:
Currently, the Keytips implementation on our documentation site isn’t working optimally because each story contains its own instance of the Keytips component. This setup was manageable when the components were only separated in stories, but it’s not effective for the docsite. Ideally, the Keytips component should be added once at the application level. With the current configuration, we would end up registering four different handlers.

### DIFF
--- a/change/@fluentui-contrib-react-keytips-a837d56c-a0f7-4c82-a6e8-4acb794f4fc7.json
+++ b/change/@fluentui-contrib-react-keytips-a837d56c-a0f7-4c82-a6e8-4acb794f4fc7.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "fix: adjust exports",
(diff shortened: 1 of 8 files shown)


dmytrokirpa:
@mainframev Could we consider implementing the `inline: false` option for KeyTips stories https://storybook.js.org/docs/7/api/doc-blocks/doc-block-story#inline?? This approach would encapsulate each story within an iframe, potentially resolving the issue of duplicate listeners.

dmytrokirpa:
After discussing with @mainframev offline, we've decided to implement a custom decorator for Keytips instead of modifying the DocsContainer. This approach ensures a single Keytips instance is used across both stories and docs mode.

dmytrokirpa (approved):

PR #255, opened 2024-10-29
feat(react-keytips): add startDelay prop

mainframev:
The feature was requested [before](https://github.com/microsoft/fluentui/issues/31500). Adding option to dely the entering in keytip mode, if start sequence is hold for a certain amount of time.


packages/react-keytips/src/hooks/useHotkeys.ts
```diff
@@ -79,15 +89,37 @@ export const useHotkeys = (hotkeys: Hotkey[], target?: Document) => {
...
+      activeKeys.current.delete(ev.key);
+
+      if (activeKeys.current.size === 0) {
+        clearDelayTimeout();
+      }
+    };
+
+    doc?.addEventListener('keydown', onKeyDown);
+    doc?.addEventListener('keyup', onKeyUp);
     return () => {
```
dmytrokirpa, line 120: Do we need to invoke the `clearDelayTimeout` function during the unmount phase?

mainframev: current keyUp ensures, that it will be called, but in general -> it's a good practice to clean all pending timeouts on umount. If this is what u mean, I can add it there explicitly

PR #256, opened 2024-10-31
fix(react-keytips): do not show keytips for disabled targets

mainframev:
According to design spec, keytips should not be shown for disabled targets

### DIFF
--- a/packages/react-keytips/package.json
+++ b/packages/react-keytips/package.json
@@ -13,7 +13,7 @@
   "dependencies": {
     "@fluentui/react-jsx-runtime": "^9.0.29",
     "@fluentui/react-utilities": "^9.16.0",
-    "@fluentui/react-positioning": "^9.16.0",
+    "@fluentui/react-positioning": "^9.15.0",
     "@fluentui/keyboard-keys": "~9.0.6",
     "@swc/helpers": "~0.5.2"
   }
--- a/packages/react-keytips/src/components/Keytips/KeytipsExamples.component-browser-spec.tsx
+++ b/packages/react-keytips/src/components/Keytips/KeytipsExamples.component-browser-spec.tsx
@@ -102,24 +102,68 @@ export const KeytipsDisabledTargetExample = (props: KeytipsProps) => {
 
   const disabledKeytipTarget = useKeytipRef({
     keySequences: ['a'],
-    content: 'A1',
+    content: 'A',
     onExecute,
   });
 
   const keytipTarget = useKeytipRef({
     keySequences: ['b'],
-    content: 'B1',
+    content: 'B',
     onExecute,
   });
 
   return (
-    <FluentProvider>
+    <>
       <Keytips {...props} />
       <Button ref={disabledKeytipTarget} disabled>
         Disabled Button
       </Button>
       <Button ref={keytipTarget}>Normal Button</Button>
-    </FluentProvider>
+    </>
+  );
+};
+
+export const KeytipsDisabledMenuTargetExample = (props: KeytipsProps) => {
+  const onExecute: ExecuteKeytipEventHandler = (_, { targetElement }) => {
+    if (targetElement) targetElement.click();
+  };
+
+  const keytipMenu = useKeytipRef({
+    keySequences: ['a'],
+    content: 'A',
+    onExecute,
+  });
+
+  const keytipMenuDisabled = useKeytipRef<HTMLDivElement>({
+    keySequences: ['a', 'b'],
+    content: 'B',
+    onExecute,
+  });
+
+  const keytipMenuEnabled = useKeytipRef<HTMLDivElement>({
+    keySequences: ['a', 'c'],
+    content: 'C',
+    onExecute,
+  });
+
+  return (
+    <>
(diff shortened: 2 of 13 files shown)


packages/react-keytips/src/components/Keytips/useKeytips.tsx
```diff
@@ -165,10 +165,10 @@ export const useKeytips_unstable = (props: KeytipsProps): KeytipsState => {
     (ev: KeyboardEvent, node: KeytipTreeNode) => {
       tree.currentKeytip.current = node;
 
-      if (node.target) {
+      if (node.target && !node.target.hasAttribute('disabled')) {
```
dmytrokirpa, line 168: For components that are neither buttons nor inputs, `aria-disabled="true"` is used instead of the `disabled` attribute. It's important to review if there are additional patterns to be aware of.

Screenshot 2024-11-01 at 14 26 49

packages/react-keytips/src/utilities/isTargetVisible.ts
```diff
@@ -1,10 +1,10 @@
-import type { PositioningProps } from '@fluentui/react-positioning';
-
 export const isTargetVisible = (
-  target?: PositioningProps['target'],
+  target?: HTMLElement,
   win?: Document['defaultView']
 ): boolean => {
   if (!target || !win) return false;
+  if (target?.hasAttribute('disabled')) return false;
```
dmytrokirpa, line 6: The same as above

packages/react-keytips/src/components/Keytips/KeytipsExamples.component-browser-spec.tsx
```diff
@@ -95,6 +95,34 @@ export const KeytipsBasicExample = (props: KeytipsProps) => {
...
+  const keytipTarget = useKeytipRef({
+    keySequences: ['b'],
+    content: 'B1',
+    onExecute,
+  });
+
+  return (
+    <FluentProvider>
+      <Keytips {...props} />
+      <Button ref={disabledKeytipTarget} disabled>
```
dmytrokirpa, line 118: Consider adding other components like Menu to the test

packages/react-keytips/package.json
```diff
@@ -13,7 +13,7 @@
   "dependencies": {
     "@fluentui/react-jsx-runtime": "^9.0.29",
     "@fluentui/react-utilities": "^9.16.0",
-    "@fluentui/react-positioning": "^9.16.0",
+    "@fluentui/react-positioning": "^9.15.0",
```
mainframev, line 16: the version was incorrect, there is no react-positioning 9.16, latest is 9.15

packages/react-keytips/src/utilities/isDisabled.ts
```diff
@@ -0,0 +1,7 @@
+export const isDisabled = (target?: HTMLElement): boolean => {
+  if (!target) return false;
+
+  return (
+    target?.hasAttribute('disabled') || target.hasAttribute('aria-disabled')
```
dmytrokirpa, line 5:
```suggestion
    target.hasAttribute('disabled') || target.hasAttribute('aria-disabled')
```

dmytrokirpa: Should we also support the `aria-disabled="false"` case?

packages/react-keytips/src/utilities/isTargetVisible.ts
```diff
@@ -1,10 +1,12 @@
...
 
 export const isTargetVisible = (
-  target?: PositioningProps['target'],
+  target?: HTMLElement,
   win?: Document['defaultView']
 ): boolean => {
   if (!target || !win) return false;
+  if (isDisabled(target)) return false;
+
   const style = win.getComputedStyle(target as HTMLElement);
```
dmytrokirpa, line 10: Shouldn't the `target` already be of the HTMLElement type?

packages/react-keytips/src/utilities/isDisabled.ts
mainframev: @dmytrokirpa I also thought about this, but currently usually it's not set explicitly to false, if it's not disabled, the aria or disabled attribute are not present. To make it more bulletproof, we can check if it's true.

packages/react-keytips/src/utilities/isTargetVisible.ts
mainframev: leftover, good catch

packages/react-keytips/src/utilities/isDisabled.ts
```diff
@@ -0,0 +1,8 @@
+export const isDisabled = (target?: HTMLElement): boolean => {
+  if (!target) return false;
+
+  return (
+    target?.hasAttribute('disabled') ||
+    target?.getAttribute('aria-disabled')?.toLowerCase() === 'true'
```
dmytrokirpa, line 6:
```suggestion
    target.hasAttribute('disabled') ||
    target.getAttribute('aria-disabled')?.toLowerCase() === 'true'
```

dmytrokirpa (approved):

PR #258, opened 2024-11-01
docs(react-keytips): improve keytips docs

mainframev:
Improving the docs for Keytips, adding more information regarding it's API


packages/react-keytips/stories/Default.stories.md
```diff
@@ -0,0 +1,4 @@
+By default Keytips are shown for all target elements that have keytip attached via
```
dmytrokirpa, line 1: Consider incorporating a jsdoc comment to describe the story component, or you could embed the description within `parameters.docs.description.story` itself, instead of using a separate markdown file. A dedicated file for a single paragraph might be unnecessarily elaborate.

packages/react-keytips/stories/OverflowMenu.stories.md
```diff
@@ -0,0 +1,14 @@
+Keytips with Overflow require `dynamic` prop to be passed with `useKeytipRef`. You can also register
```
dmytrokirpa, line 1: While the pattern is indeed used in the fluentui core, the file is not a Storybook story. Hence, it should be named `OverflowMenu.md` instead of `OverflowMenu.stories.md`

dmytrokirpa (approved):

PR #260, opened 2024-11-21
feat(react-data-grid): support autoFitColumns: false

### LINKED ISSUE #176: [Feat] Virtualized DataGrid with autoFitColumns: false
For the virtualized dataGrid (@fluentui-contrib/react-data-grid-react-window), the scrollbar is on the body

This is fine if the columns all fit without a horizontal overflow, however if there is some horizontal scrollbar there's no way to scroll both the body and the headers at the same time. So when `autoFitColumns` is set to false, this causes issues

`overflow-x: auto` can be applied to the parent container, but that means the vertical scrollbar won't be visible unless the user scrolls all the way to the right
image
https://codesandbox.io/p/sandbox/datagrid-body-overflow-forked-ygzvhs?workspaceId=2e8388ff-7617-4f14-8fb7-fe6621520941

`overflow-x: auto` can be applied to the `DataGridBody`, but that means the `DataGridHeader`s won't scroll with the body and the column data and headers will be misaligned
image
https://codesandbox.io/p/sandbox/datagrid-bodysize-bug-forked-6xqrry?file=%2Fsrc%2Fexample.tsx

(for the sandboxes, try resizing some columns to be larger and then observe the scrollbar positions/behaviours)

rocketBANG:
* Adds some onScroll events to both the body and the header to allow them to stay in sync even though each has their own scroll bar.
  * Previously, when using resizableColumns and resizableColumnsOptions.autoFitColumns = false, only the body or the header could be scrolled at once and they would be out of sync
  * This changes synchronizes the scrolling between the two so scrolling horizontally on either affects the other as if you were scrolling on a shared parent
* Also updated some styling to support as well
* Added a story to showcase the support

Fixes #176
Fixes https://github.com/microsoft/fluentui/issues/32837


packages/react-data-grid-react-window/src/components/DataGrid/renderDataGrid.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+/**
+ * Render the final JSX of DataGrid
+ */
+export const renderDataGrid_unstable = (
+  state: DataGridState,
+  contextValues: DataGridContextValues
+) => {
+  const headerRef = React.useRef<HTMLDivElement | null>(null);
+  const bodyRef = React.useRef<HTMLDivElement | null>(null);
```
ling1726, line 18: Please use hooks in the state hook (i.e. `useDataGrid` hook) everything inside a render function should be a pure function

packages/react-data-grid-react-window/src/components/DataGridBody/useDataGridBody.tsx
```diff
@@ -54,6 +59,28 @@ export const useDataGridBody_unstable = (
...
+  }, []);
+
+  // Use a onScroll callback on the outerElement since react-window's onScroll will only work for horizontal scrolls
+  const setupOuterRef = React.useCallback(
+    (outerElement: HTMLElement | null) => {
+      bodyRef.current?.removeEventListener('scroll', onScroll);
+
+      bodyRef.current = outerElement;
+      if (outerElement) {
+        outerElement.addEventListener('scroll', onScroll);
```
ling1726, line 78: This event listener on the other hand needs to be removed on ref disposal. Easiest way to do this is to return a callback ref from `useMemo` so that you can keep a handle to the element on ref dispose - smth along the lines of this...

```ts
const ref = React.useMemo(() => {
  let dispose = () => null;

  return (el) => {
      if (el) {
        el.addEventListener();
        dispose = () => el.removeEventListener()
      } else {
        dispose();
      }
   }
})
```

packages/react-data-grid-react-window/src/components/DataGridHeader/useDataGridHeader.ts
```diff
@@ -0,0 +1,58 @@
...
+  const onScroll = React.useCallback(() => {
+    if (bodyRef.current && headerRef.current) {
+      bodyRef.current.scroll({
+        left: headerRef.current.scrollLeft,
+        behavior: 'instant',
+      });
+    }
+  }, []);
+
+  const setupRef = React.useCallback((element: HTMLElement | null) => {
```
ling1726, line 43: you do not need this `setRef`, because Fluent UI already has a utility for this, you can use it like so

```ts
  const setupRef = React.useCallback((element: HTMLElement | null) => {
    headerRef.current?.removeEventListener('scroll', onScroll);

    headerRef.current = element;
    if (element) {
      element.addEventListener('scroll', onScroll);
    }
  }, []);

  const baseState = useBaseState(props, setupRef);

  baseState.root.ref = useMergedRefs(baseState.root.ref, setupRef);

  return baseState;
};
```

packages/react-data-grid-react-window/src/components/DataGrid/useDataGridStyles.styles.ts
```diff
@@ -0,0 +1,27 @@
...
+  makeStyles,
+  mergeClasses,
+  useDataGridStyles_unstable as useDataGridStylesBase_unstable,
+} from '@fluentui/react-components';
+
+const useStyles = makeStyles({
+  root: {
+    // DataGrid gets min-width: fit-content applied directly to the element, thus the need for !important
+    // without auto width, the DataGrid will not scroll
+    minWidth: 'auto !important',
```
ling1726, line 12: instead of `!important` styles that are applied in all cases, how about just override the inline style from the component's state in the resizable columns scenario?

```ts
export const useDataGrid_unstable = (
  props: DataGridProps,
  ref: React.Ref<HTMLElement>
): DataGridState => {
  const { targetDocument } = useFluent();
  const scrollbarWidth = useScrollbarWidth({ targetDocument });

  let containerWidthOffset = props.containerWidthOffset;

  if (containerWidthOffset === undefined) {
    containerWidthOffset = props.selectionMode
      ? -TABLE_SELECTION_CELL_WIDTH
      : 0;
    containerWidthOffset -= scrollbarWidth || 0;
  }

  const baseState = useBaseState(
    { ...props, 'aria-rowcount': props.items.length, containerWidthOffset },
    ref
  );

  if (props.resizableColumns && baseState.root.style) {
    baseState.root.style.minWidth = 'auto';
  }

  return baseState;
};

```

packages/react-data-grid-react-window/src/components/DataGrid/renderDataGrid.tsx
rocketBANG: Ah I didn't realize it should be pure, updated now

packages/react-data-grid-react-window/src/components/DataGridHeader/useDataGridHeader.ts
rocketBANG: That's much better - updated to use the utility

packages/react-data-grid-react-window/src/components/DataGrid/useDataGridStyles.styles.ts
rocketBANG: yep that makes much more sense, a lot better than needing `!important` - updated

ling1726 (approved):

PR #263, opened 2024-11-21
chore(react-tree-grid): updates LiveMeetingsList story for better navigation

bsunderhus:
Updates `LiveMeetingsList` stories to ensure a better narration.

1. move each information provided on a row for it's own gridcell

### DIFF
--- a/packages/react-tree-grid/stories/LiveMeetingsList.stories.tsx
+++ b/packages/react-tree-grid/stories/LiveMeetingsList.stories.tsx
@@ -270,25 +270,19 @@ const MeetingsSectionItem = (props: MeetingsSectionItemProps) => {
           </Tag>
         )}
       </TreeGridCell>
-      {props.tasks && (
-        <TreeGridCell className={styles.tasks}>{props.tasks}</TreeGridCell>
-      )}
-      {props.attachments && (
-        <TreeGridCell className={styles.attachments}>
-          {props.attachments}
-        </TreeGridCell>
-      )}
+      <TreeGridCell className={styles.tasks}>{props.tasks}</TreeGridCell>
+      <TreeGridCell className={styles.attachments}>
+        {props.attachments}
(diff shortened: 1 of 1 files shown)


packages/react-tree-grid/stories/LiveMeetingsList.stories.tsx
```diff
@@ -267,49 +243,45 @@ const MeetingsSectionItem = (props: MeetingsSectionItemProps) => {
...
-      {props.attachments && <TreeGridCell>{props.attachments}</TreeGridCell>}
-      <TreeGridCell>
+      {props.tasks && (
+        <TreeGridCell className={styles.tasks}>{props.tasks}</TreeGridCell>
+      )}
+      {props.attachments && (
+        <TreeGridCell className={styles.attachments}>
+          {props.attachments}
+        </TreeGridCell>
+      )}
```
ling1726, line 280: what is the conclusion on assymetric rows for treegrids? should we have them or not?

bsunderhus: There's no conclusion, we're focusing at the moment on the proper narration when you navigate on a nested level (from row to cells).

I believe the asymmetry might be an issue when navigating between rows, on a vertical level, we still need to verify that with more screen readers.

kolaps33: Good point I think would be better to render just empty gridcell. 

Because when trying to navigation with NVDA in column navigation ctrl+alt+up/down arrows, then when there is asymmetry, then screen reader jump to different column automatically. 

This is good point how to solve it properly :)

bsunderhus: Ok, just updated this, now it should be symmetric in this example 👍🏻

PR #269, opened 2024-12-10
 refactor (react-data-grid-react-window): export all types and hooks git committed these change files

### LINKED ISSUE #265: Issues making variant of Virtualized DataGrid
Is it possible to make a variant of the DataGrid in this repo? I am trying to utilize this DataGrid as it does have virtualization but also add more features on top of it but it doesn't look like this library exposes access to the renderDataGrid_unstable function nor the context providers in this repo making it difficult to make edits internally unless we copy/paste a lot of the code within the repo which i would like to avoid. Are there any suggestions here?

alton-browsing:
Export all hooks and types to make it easier for consumers to extend the control. Also fix the naming of the DataGrid.types.ts file which was wrongly named as DataGridBody.types.ts before

Fixes #265


ling1726 (approved):

ling1726:
@alton-browsing can you please fix formatting errors in code? you should be able to run `yarn nx format`

alton-browsing:
@ling1726 should be fixed now

ling1726 (approved):

PR #270, opened 2024-12-11
feat(react-keytips): add support for menu shorcuts

mainframev:
This feature aims to provide support for previously available 'persisted' keytips in v8. In v8, the [OverflowSet](https://github.com/microsoft/fluentui/blob/master/packages/react/src/components/OverflowSet/OverflowButton.tsx) was responsible for registering persisted keytips inside. 

The documentation and tests will be added as follow up to this PR


packages/react-keytips/src/constants.ts
```diff
@@ -13,12 +12,15 @@ export const KTP_SEPARATOR = '-';
...
-export const SHOW_DELAY = 250;
+export const SHOW_DELAY = 30;
+export const INVISIBLE_KEYTIPS_ID = 'invisible-keytips-wrapper';
 
 export const EVENTS = {
   KEYTIP_ADDED: 'fui-keytip-added',
   KEYTIP_REMOVED: 'fui-keytip-removed',
   KEYTIP_UPDATED: 'fui-keytip-updated',
+  KEYTIP_EXECUTED: 'fui-keytip-executed',
+  SHORTCUT_EXECUTED: 'fui-shortcut-executed',
```
dmytrokirpa, line 23: nit: would be bice to prefix all event names with fui-keytip-* for consistency

packages/react-keytips/src/hooks/useTree.ts
```diff
@@ -155,15 +175,19 @@ export function useTree() {
...
-    const fullSequence = keytip.keySequences.slice(0, -1);
+    const sequence = keytip.keySequences.slice(0, -1) ?? [];
     const parentID =
-      fullSequence.length === 0 ? KTP_ROOT_ID : sequencesToID(fullSequence);
+      sequence.length === 0 ? KTP_ROOT_ID : sequencesToID(sequence);
     return currentKeytip.current.id === parentID;
   }, []);
 
+  const getNode = (id: string) =>
+    [...nodeMap.current.values()].find((node) => node.id === id);
```
dmytrokirpa, line 185: I know that JS Set doesn't have .find, and for-of might be more verbose, but it will decrease complexity from 2N to N

dmytrokirpa (approved):
Left a few minor comments, otherwise LGTM!

PR #275, opened 2025-01-06
test(react-keytips): add tests for shortcut

mainframev:
A few tests for Keytip shortcuts.

### DIFF
--- a/change/@fluentui-contrib-react-keytips-c341fb03-d4b1-4cd3-9f39-851d2246ea52.json
+++ b/change/@fluentui-contrib-react-keytips-c341fb03-d4b1-4cd3-9f39-851d2246ea52.json
@@ -1,0 +1,7 @@
+{
+  "type": "none",
+  "comment": "test: add test for shortcut keytip",
+  "packageName": "@fluentui-contrib/react-keytips",
+  "email": "vgenaev@gmail.com",
+  "dependentChangeType": "none"
+}
--- a/packages/react-keytips/src/components/Keytips/Keytips.component-browser-spec.tsx
+++ b/packages/react-keytips/src/components/Keytips/Keytips.component-browser-spec.tsx
@@ -119,7 +119,7 @@ test.describe('test keytip navigation', () => {
 
     await expect(page.getByRole('tooltip', { name: 'A' })).toBeVisible();
     // should open nested menus and show next keytip
-    await page.keyboard.press('a');
+    await component.press('a');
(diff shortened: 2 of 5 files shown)


package.json
```diff
@@ -47,7 +47,7 @@
     "@nx/web": "19.3.2",
     "@nx/webpack": "19.3.2",
     "@nx/workspace": "19.3.2",
-    "@playwright/experimental-ct-react": "^1.41.2",
+    "@playwright/experimental-ct-react": "^1.49.1",
```
mainframev, line 50: had to update the playwright because of failing browsers installation, similar as [here](https://github.com/microsoft/playwright/issues/33051)

packages/nx-plugin/src/executors/playwright/executor.ts
```diff
@@ -114,7 +115,8 @@ function createArgs(
  */
 function runPlaywright(args: string[], cwd: string) {
   try {
-    const cli = require.resolve('@playwright/experimental-ct-core/cli');
+    const module = require.resolve('@playwright/experimental-ct-react');
```
mainframev, line 118: `cli` module isn't part of package.json exports, so we cannot access it as before anymore

.github/workflows/ci.yml
```diff
@@ -45,7 +45,7 @@ jobs:
       - run: yarn nx affected --target=type-check
       - run: yarn nx affected --target=lint
       - run: yarn nx affected --target=test
-      - run: yarn nx affected --target=component-test --skipInstall
+      - run: yarn nx affected --target=component-test --skipInstall --parallel --max-parallel=1
```
mainframev, line 48: @hotell

Looks like playwright is not close the server after each target tested, when running with `nx affected`.

Playwright first reports: `Dev Server is already running at http://localhost:3100/, using it`, later fails with several errors: "Error: Could not connect to the server", "Error: Unregistered component". Happens locally as well with `nx affected`.

package.json
Hotell: can we handle these things in separate PR, this is unrelated to the scope of this PR.

.github/workflows/ci.yml
Hotell, line 48:
```suggestion
      - run: yarn nx affected --target=component-test --skipInstall --parallel 1
```

Hotell (commented):
pls move playwright related changes to separate PR. ty

dmytrokirpa (approved):

PR #276, opened 2025-01-07
chore: update playwright to 1.49.1

mainframev:


Hotell:
seems the error is still present even if you use parallel 1 

https://github.com/microsoft/fluentui-contrib/actions/runs/12667454076/job/35300855387?pr=276

.github/workflows/ci.yml
```diff
@@ -45,7 +45,8 @@ jobs:
       - run: yarn nx affected --target=type-check
       - run: yarn nx affected --target=lint
       - run: yarn nx affected --target=test
-      - run: yarn nx affected --target=component-test --skipInstall
+      - run: npx playwright clear-cache
```
Hotell, line 48: why is this needed ?

Hotell:
I did further investigation. this solution will not work unfortunately.


we need to dynamically set `ctPort` on every package to get deterministic port opnenings on both local and dev

see https://github.com/microsoft/fluentui-contrib/pull/278

the manual steps are the only option ATM, once we migrate contrib to use inferred targets this will be handled with less maintenance.

.github/workflows/ci.yml
mainframev: ignore this, I was checking why it's still failing, as locally it was fine and only on CI failing. I guess I can close this PR, let's use yours with the fix.

mainframev:
closing in favour #278

.github/workflows/ci.yml
Hotell: kk sounds good. ty

PR #278, opened 2025-01-08
chore: update playwright to 1.49.1 and resolve CT port clashes on ci/local

Hotell:
replaces https://github.com/microsoft/fluentui-contrib/pull/276

### DIFF
--- a/packages/nx-plugin/src/generators/playwright-component-configuration/files/playwright.config.ts.template
+++ b/packages/nx-plugin/src/generators/playwright-component-configuration/files/playwright.config.ts.template
@@ -33,7 +33,8 @@ export default defineConfig({
     trace: 'on-first-retry',
 
     /* Port to use for Playwright component endpoint. */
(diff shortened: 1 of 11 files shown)


packages/nx-plugin/src/generators/playwright-component-configuration/files/playwright.config.ts.template
```diff
@@ -33,7 +33,8 @@ export default defineConfig({
     trace: 'on-first-retry',
 
     /* Port to use for Playwright component endpoint. */
-    ctPort: 3100,
+    // eslint-disable-next-line @typescript-eslint/no-var-requires
+    ctPort: 3100 + Number(require('./package.json').name.length),
```
Hotell, line 37: this is only way to avoid port clashes when running multiple ct test both localy and on CI. 

once we migrate to nx inferred projects this will be generated behind the scenes

mainframev (approved):

PR #284, opened 2025-01-15
chore: fix invalid project configurations

Hotell:
see PR title


packages/react-themeless-provider/tsconfig.json
```diff
@@ -13,7 +13,7 @@
       "path": "./tsconfig.spec.json"
     },
     {
-      "path": "./tsconfig.storybook.json"
+      "path": "./.storybook/tsconfig.json"
```
Hotell, line 16: TS was not working for stories

packages/react-themeless-provider/package.json
```diff
@@ -8,10 +8,6 @@
     "@griffel/react": "^1.5.14",
     "@swc/helpers": "~0.5.2"
   },
-  "devDependencies": {
```
Hotell, line 11: we dont specify devDeps ( everything is consumed from monorepo root ) - single version policy

packages/azure-theme/package-lock.json
Hotell, line 1: only 1 lock file exists in monorepo - this probably slip through PR reviews

layershifter (approved):

mainframev (approved):

PR #288, opened 2025-01-17
fix(react-keytips): do not show overflowing keytips

mainframev:
**Before:**

Screenshot 2025-01-17 at 19 50 04

**After:**

Screenshot 2025-01-17 at 19 51 04

In the Overflow example, the overflowed element stays in the DOM after it’s overflowed. Just checking the visibility or display styles isn’t enough. For example, with a Keytip attached to a SplitButton, the styles were applied to the root div of the SplitButton, not the specific side button where the Keytip was attached. To make this more reliable, we could add a check for the data-overflowing attribute.

### DIFF
--- a/packages/react-keytips/src/utilities/isTargetVisible.ts
+++ b/packages/react-keytips/src/utilities/isTargetVisible.ts
@@ -6,7 +6,12 @@ export const isTargetVisible = (
 ): boolean => {
   if (!target || !win) return false;
   if (isDisabled(target)) return false;
+  if (
+    target?.hasAttribute('data-overflowing') ||
+    target?.parentElement?.hasAttribute('data-overflowing')
+  )
+    return false;
 
   const style = win.getComputedStyle(target);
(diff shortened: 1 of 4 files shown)


packages/react-keytips/src/utilities/isTargetVisible.ts
```diff
@@ -6,7 +6,8 @@ export const isTargetVisible = (
 ): boolean => {
   if (!target || !win) return false;
   if (isDisabled(target)) return false;
+  if (target?.parentElement?.hasAttribute('[data-overflowing]')) return false;
```
mainframev, line 9: [checkVisibility](https://developer.mozilla.org/en-US/docs/Web/API/Element/checkVisibility) could be a better option, if not our browser matrix support and jsDOM not supporting it [yet](https://github.com/jsdom/jsdom/issues/3695) 😞

ValentinaKozlova (approved):

PR #281, opened 2025-01-13
Migrate to version plans

Hotell:

PR #292, opened 2025-01-20
chore: setup nx release with version plans

Hotell:
Adds production configuration for Nx Release Plans and publishing.

> see https://nx.dev/recipes/nx-release/file-based-versioning-version-plans#file-based-versioning-version-plans

Relates to https://github.com/microsoft/fluentui-contrib/pull/281

### DIFF
--- a/nx.json
+++ b/nx.json
@@ -37,6 +37,11 @@
     },
     "e2e": {
       "cache": true
+    },
+    "nx-release-publish": {
+      "options": {
+        "packageRoot": "dist/packages/{projectName}"
+      }
     }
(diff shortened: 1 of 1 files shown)


nx.json
```diff
@@ -53,6 +58,43 @@
     ],
     "sharedGlobals": []
   },
+  "release": {
+    "projects": ["packages/**/*", "!tag:npm:private"],
```
Hotell, line 62: any non private project within packages/ is publishable

mainframev (approved):

PR #308, opened 2025-02-11
feat: Add Button components for teams-components package

ling1726:
Adds strict API for Button, ToggleButton, MenuButton components that:

* restricts appearance
* supports and enforces tooltip for icon buttons
* enforces strics CSS classes
* supports strict icon and toltip slots with lightweight types that are compatible with Fluent UI slots

### DIFF
--- a/packages/teams-components/src/components/Button/Button.tsx
+++ b/packages/teams-components/src/components/Button/Button.tsx
@@ -25,7 +25,7 @@ export interface ButtonProps
   className?: StrictCssClass;
   icon?: StrictSlot;
   onClick?: React.MouseEventHandler<HTMLButtonElement>;
-  tooltip?: StrictSlot;
+  title?: StrictSlot;
 }
 
 export const Button = React.forwardRef<HTMLButtonElement, ButtonProps>(
@@ -34,7 +34,7 @@ export const Button = React.forwardRef<HTMLButtonElement, ButtonProps>(
       validateProps(userProps);
     }
 
-    const { className, icon, tooltip, ...restProps } = userProps;
+    const { className, icon, title, ...restProps } = userProps;
     const props: ButtonPropsBase = {
       ...restProps,
       className: className?.toString(),
@@ -47,9 +47,9 @@ export const Button = React.forwardRef<HTMLButtonElement, ButtonProps>(
 
     const button = renderButton_unstable(state);
 
-    if (tooltip) {
+    if (title) {
       return (
-        <Tooltip content={tooltip} relationship="label">
+        <Tooltip content={title} relationship="label">
           {button}
         </Tooltip>
       );
--- a/packages/teams-components/src/components/Button/Button.tsx
+++ b/packages/teams-components/src/components/Button/Button.tsx
@@ -6,6 +6,7 @@ import {
   type ButtonProps as ButtonPropsBase,
   Tooltip,
 } from '@fluentui/react-components';
+import { validateIconButton } from './validateIconButton';
 import { type StrictCssClass, validateStrictClasses } from '../../strictStyles';
 import { type StrictSlot } from '../../strictSlot';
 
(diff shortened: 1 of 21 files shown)


packages/teams-components/src/components/ToggleButton/ToggleButton.tsx
```diff
@@ -0,0 +1,67 @@
...
+    | 'children'
+    | 'disabled'
+    | 'disabledFocusable'
+  > {
+  appearance?: 'transparent' | 'primary';
+  className?: StrictCssClass;
+  icon?: StrictSlot;
+  onClick?: React.MouseEventHandler<HTMLButtonElement>;
+  tooltip?: StrictSlot;
+  checked: boolean;
```
layershifter, line 29: `onChange` too, then?

packages/teams-components/src/components/ToggleButton/ToggleButton.tsx
```diff
@@ -0,0 +1,67 @@
...
+  useToggleButtonStyles_unstable,
+  useToggleButton_unstable,
+  renderToggleButton_unstable,
+  Tooltip,
+} from '@fluentui/react-components';
+import { validateStrictClasses, type StrictCssClass } from '../../strictStyles';
+import type { StrictSlot } from '../../strictSlot';
+import { validateIconButton } from '../Button/validateIconButton';
+
+export interface ToggleButtonProps
```
layershifter, line 13: Should it reuse `ButtonProps`?

packages/teams-components/src/components/Button/validateIconButton.ts
```diff
@@ -0,0 +1,21 @@
...
+  icon?: unknown;
+  children?: unknown;
+  tooltip?: unknown;
+  'aria-label'?: string;
+  'aria-labelledby'?: string;
+}) => {
+  if (
+    !props.children &&
+    props.icon &&
+    !props.tooltip &&
```
layershifter, line 14: Should be this `Tooltip` or `title`?

packages/teams-components/src/components/Button/Button.tsx
```diff
@@ -0,0 +1,65 @@
...
+    };
+
+    let state = useButton_unstable(props, ref);
+    state = useButtonStyles_unstable(state);
+
+    const button = renderButton_unstable(state);
+
+    if (tooltip) {
+      return (
+        <Tooltip content={tooltip} relationship="label">
```
layershifter, line 52: We discussed to have only controlled components: it would be a funny problem if somebody wants to control this `Tooltip`.

packages/teams-components/src/components/ToggleButton/ToggleButton.tsx
ling1726: There is onChange for this component 🙃

packages/teams-components/src/components/ToggleButton/ToggleButton.tsx
ling1726: Yes I believe some parts of this should be shared, let me check

packages/teams-components/src/components/Button/Button.tsx
ling1726: hmm let's call it `title` them as you suggested, to remove the notion that it's a tooltip that can be controlled - people can always compose a Tooltip and and use the title prop

packages/teams-components/src/components/Button/validateIconButton.ts
ling1726: let's stick with title - I'm gonna go with the thought that our tooltip for buttons is basically an enhanced DOM title

packages/teams-components/src/components/ToggleButton/ToggleButton.tsx
ling1726: Yeah just making everything extend the strict button props makes much more sense

PR #310, opened 2025-02-14
feat: Extensions to makeStrictStyles

ling1726:
Updates strict styles API to support component level extensions by leveraging generics throughout all the types. Example of usage

```tsx
const useStrictStyles = makeStrictStyles<typeof Button>({
  strict: {
    alignSelf: 'center',
    // 👇 `maxWidth` only exists when Button style extensions are used
    maxWidth: '100%',
  },
});

// 👇 This is equally valid usage for the `Button` component without enabling component styling extensions
const useStrictStyles = makeStrictStyles({
  strict: {
    alignSelf: 'center',
  },
});

<Button className={strictStyles.strict}>Button</Button>
```

### DIFF
--- a/packages/teams-components/src/strictStyles/types.ts
+++ b/packages/teams-components/src/strictStyles/types.ts
@@ -1,18 +1,20 @@
+import * as React from 'react';
 import type { STRICT_SYMBOL } from './STRICT_SYMBOL';
 import type { GriffelStyle } from '@fluentui/react-components';
 
-export interface StrictCssClass {
+export interface StrictCssClass<TExtension = DefaultStrictStyles> {
   /**
    * @returns CSS class string
    */
   toString: () => string;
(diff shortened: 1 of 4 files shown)


packages/teams-components/src/strictStyles/types.ts
```diff
@@ -1,18 +1,20 @@
...
+export interface StrictCssClass<TExtension = DefaultStrictStyles> {
   /**
    * @returns CSS class string
    */
   toString: () => string;
   DO_NOT_USE_OR_YOU_WILL_BE_FIRED: typeof STRICT_SYMBOL;
+  /**
+   * This field is only used to allow style extensions for components
+   */
+  overrideFilter?: TExtension;
```
bsunderhus, line 14: nit: you could use a symbol to hide this property

bsunderhus (approved):

layershifter (approved):

PR #311, opened 2025-02-18
feat: Add Toolbar component to teams-components

ling1726:


packages/teams-components/src/components/Toolbar/ToolbarDivider.tsx
```diff
@@ -0,0 +1,30 @@
...
+  const styles = useStyles();
+  const state = useDivider_unstable(
+    {
+      ...props,
+      vertical: true,
+      className: mergeClasses(toolbarDividerClassNames.root, styles.root),
+    },
+    ref
+  );
+  useDividerStyles_unstable(state);
```
layershifter, line 27: I would double check if we really want to reuse `Divider` there

packages/teams-components/src/components/Toolbar/index.ts
```diff
@@ -0,0 +1,5 @@
+export * from './Toolbar';
+export * from './ToolbarDivider';
+export * from './ToolbarButton';
+export * from './ToolbarToggleButton';
+export * from './ToolbarMenuButton';
```
layershifter, line 5: This is too wild, let's be explicit

packages/teams-components/stories/Toolbar/Default.stories.tsx
```diff
@@ -0,0 +1,184 @@
...
+          </ToolbarMenuButton>
+        </MenuTrigger>
+        <MenuPopover>
+          <MenuList>
+            <MenuItem>Max Mustermann</MenuItem>
+            <MenuItem>Jane Doe</MenuItem>
+            <MenuItem>Pierre Dupont</MenuItem>
+          </MenuList>
+        </MenuPopover>
+      </Menu>
```
layershifter, line 150: This one is interesting: what should we do with `Menu`?

packages/teams-components/src/elementWalker.ts
```diff
@@ -0,0 +1,70 @@
+import { isHTMLElement } from '@fluentui/react-utilities';
+
+type ElementFilter = (node: HTMLElement) => number;
+
+export class HTMLElementWalker {
+  private _treeWalker: TreeWalker;
+  private _filter: ElementFilter;
```
layershifter, line 7:
```suggestion
  #treeWalker: TreeWalker;
  #filter: ElementFilter;
```

let's be cool

packages/teams-components/src/components/Toolbar/Toolbar.tsx
```diff
@@ -0,0 +1,168 @@
...
+      !isAllowedToolbarItem(child) &&
+      !isPortalSpan(child) &&
+      !isTabsterDummy(child)
+    ) {
+      throw new Error(
+        '@fluentui-contrib/teams-components::Toolbar::Use Toolbar components from @fluentui-contrib/teams-components package only'
+      );
+    }
+  }
+};
```
layershifter, line 141: We will definitely need a nicer validator

packages/teams-components/src/components/Toolbar/Toolbar.tsx
```diff
@@ -0,0 +1,168 @@
...
+      validateToolbarItems(elRef.current);
+    }
+
+    const treeWalker = new HTMLElementWalker(elRef.current, (el) => {
+      if (isAllowedToolbarItem(el) || el === treeWalker.root) {
+        return NodeFilter.FILTER_ACCEPT;
+      }
+
+      return NodeFilter.FILTER_REJECT;
+    });
```
layershifter, line 67: TreeWalker is nice, but why? `element.children` with a filter will do the same

packages/teams-components/src/components/Toolbar/Toolbar.tsx
```diff
@@ -0,0 +1,168 @@
...
+  }
+
+  if (el.classList.contains(toolbarDividerClassNames.root)) {
+    el.style.marginInlineStart = tokens.spacingHorizontalS;
+    return;
+  }
+
+  treeWalker.currentElement = el;
+  const prev = treeWalker.previousElement();
+  if (prev && prev.dataset.appearance !== 'transparent') {
```
layershifter, line 159: I don't think that relying on DOM is a good idea. Registration would work better: we don't expose anything to DOM, we are free to store structure of any complexity

packages/teams-components/src/components/Toolbar/Toolbar.tsx
```diff
@@ -0,0 +1,168 @@
...
+    el.style.marginInlineStart = tokens.spacingHorizontalS;
+    return;
+  }
+
+  if (prev && el.dataset.appearance !== 'transparent') {
+    prev.style.marginInlineStart = tokens.spacingHorizontalS;
```
layershifter, line 165: Have to say, that this should be somehow declarative as it will go will quite quickly. 

Dumb idea:

```
const STYLE_SCHEMAS = {
  kind: 'ToolbarButton',
  prevKind: 'ToolbarDivider',
  nextKind: null,
  meta: { appearance },
  style: { marginInlineStart: tokens.spacingHorizontalS },
}

function applyStylesToElement() {}
```

packages/teams-components/src/components/Toolbar/Toolbar.tsx
ling1726: will try out registration and compare

packages/teams-components/src/components/Toolbar/Toolbar.tsx
ling1726: As discussed offline - will try out registerProperty API without inheritable

ling1726: we could try leverage default react context or other means to make sure the cost of property registration only happens with usage

packages/teams-components/stories/Toolbar/Default.stories.tsx
ling1726: We have plans for strict menu, resolving this for now since we'll get back to this

packages/teams-components/src/components/Toolbar/itemRegistration.ts
```diff
@@ -0,0 +1,123 @@
...
+    recalcToolbarItemSpacing();
+    firstMount.current = false;
+  }, [recalcToolbarItemSpacing, targetDocument]);
+
+  return registerItem;
+};
+
+export const useItemRegistration = (metadata: ToolbarItemMetadata) => {
+  const { registerItem } = React.useContext(ToolbarItemRegistrationContext);
+  const styles = useItemRegistrationStyles();
```
layershifter, line 115: not sure about mixing styles there

packages/teams-components/src/components/Toolbar/itemRegistration.styles.ts
```diff
@@ -0,0 +1,32 @@
+import { makeStrictStyles } from '../../strictStyles/makeStrictStyles';
+
+export const itemRegistrationVars = {
+  toolbarItemMarginInlineStart: '--toolbar-item-margin-inline-start',
+};
+
+let propertyRegisterComplete = false;
+
+export const registerCustomProperties = (win: typeof globalThis) => {
+  if (propertyRegisterComplete) {
```
layershifter, line 10: Should be done per window?

packages/teams-components/src/components/Toolbar/itemRegistration.ts
```diff
@@ -0,0 +1,123 @@
...
+  const ref = React.useCallback((el: HTMLElement) => {
+    if (el) {
+      registerItem(el, metadata);
```
layershifter, line 118: We don't react on `meta` updates

PR #314, opened 2025-02-21
chore(ci): update github pages upload action

mainframev:
Failed job: https://github.com/microsoft/fluentui-contrib/actions/runs/13457302489 

`actions/upload-artifact: v3` is [no longer accessible for usage](https://github.blog/changelog/2024-04-16-deprecation-notice-v3-of-the-artifact-actions/). 

Upload to the latest [upload-page-artifact](https://github.com/actions/upload-pages-artifact) should solve it: https://github.com/actions/upload-pages-artifact/blob/main/action.yml#L80

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -43,12 +43,12 @@ jobs:
       - run: yarn nx run-many --target=build-storybook
 
       - name: Setup Pages
-        uses: actions/configure-pages@v3
+        uses: actions/configure-pages@v5
       - name: Upload artifact
-        uses: actions/upload-pages-artifact@v1
+        uses: actions/upload-pages-artifact@v3
         with:
           # Upload entire repository
           path: './dist/storybook'
       - name: Deploy to GitHub Pages
         id: deployment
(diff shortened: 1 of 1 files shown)


TristanWatanabe (approved):

.github/workflows/docs.yml
```diff
@@ -10,6 +10,7 @@ permissions:
   contents: read
   pages: write
   id-token: write
+  actions: read
```
Hotell, line 13: why is this needed ?

Hotell: only access added should be following ( am I missing something ? ) https://github.com/microsoft/fluentui/pull/33829/files#diff-0e99d05f4114307b51d9215a468223cb3314c669b1a8add474d5468ccb9337ebL60-L62

mainframev: @Hotell hm, I read it from in changelog note [here](https://github.com/actions/deploy-pages/releases/tag/v4.0.0). But if it works without, then I will remove it.

PR #321, opened 2025-02-28
fix(react-gamepad-navigation): remove enums and native events augmentation

hectorjjb:
Follow up PR to address items below:

- Ensure the package isn't using enums
- modify useGamepandNavigation hook so its logic in under React.useEffect (React.useMemo for performance)
- hange handleSelect* functions signatures to ensure they are only called on select elements
- Remove augment properties on mouse & keyboard events


hectorjjb:
@dmytrokirpa @marcosmoura follow-up PR as discussed previously

marcosmoura (approved):

PR #336, opened 2025-03-20
Update "@fluentui/react-icons" dependency

adr1-jpg:
- Removed "@fluentui/react-icons" as a dependency. It is only being used by the storybook components, not the package itself. This is to ensure that there are no issues/errors locally if there is a mismatch between the versions. 

- Added @mhdahman as package owner for azure-theme


mainframev (approved):

Hotell:
@adr1-jpg we have now tooling in place that would mitigate these kind of dependency misaligments 

- you can undo the package changes which are handled by https://github.com/microsoft/fluentui-contrib/pull/336
- keep the team ownership change 

once fixed we can merge. ty!

PR #349, opened 2025-04-01
Scaffolding token-analyzer package

brandonthomas:
PR to add the token-analyzer package into the contrib repo. See https://github.com/microsoft/fluentui/pull/34108 for a preview of logic. That PR will be abandoned as we're moving the code into contrib.


brandonthomas:
microsoft/fluentui#34108

nx.json
```diff
@@ -38,6 +38,11 @@
       "options": {
         "packageRoot": "dist/packages/{projectName}"
       }
+    },
+    "@nx/js:swc": {
```
dmytrokirpa, line 42: I know that this was added by generator, but please revert it

package.json
```diff
@@ -99,8 +99,8 @@
     "syncpack": "^9.8.6",
     "ts-node": "10.9.1",
     "tslib": "^2.3.0",
-    "typescript-eslint": "8.25.0",
     "typescript": "5.7.3",
+    "typescript-eslint": "8.25.0",
```
dmytrokirpa, line 103: This is likely not related to the `token-analyzer` scaffolding, so we might need to revert it as well.

brandonthomas: I didn't add this either. The scaffold script does ask what lint I want to use so I assumed this was part of that.

nx.json
brandonthomas: We should open an issue to fix this then.  As someone who hasn't contributed here before, I would have no way of knowing what is or isn't erroneously done by the generator.

package.json
brandonthomas: https://github.com/microsoft/fluentui-contrib/issues/350

brandonthomas: I'm wondering if this happened from a yarn install or something? It's just a reorder.

PR #352, opened 2025-04-07
feat(react-contextual-pane): project bootstrap

chpalac:
## Overview 

Contextual Pane project scaffolding


nx.json
```diff
@@ -38,6 +38,11 @@
       "options": {
         "packageRoot": "dist/packages/{projectName}"
       }
+    },
+    "@nx/js:swc": {
```
Hotell, line 42: plsease remove this one - guess it's a bug in our generator. ty

PR #372, opened 2025-05-09
feat(resize-handle): support viewport units

layershifter:
This PR adds support for viewport units i.e. `vh` & `vw` for `useResizeHandle()` via an option:

```ts
  const { handleRef, wrapperRef, elementRef } = useResizeHandle({
    variableName: '--width',
    growDirection: 'end',
    unit: 'viewport',
  });
```

The option defaults to pixels (`px`). Callbacks will return also a used unit:

```ts
  const { handleRef, wrapperRef, elementRef } = useResizeHandle({
    unit: 'viewport',
    onChange: (ev, data) => {
      data.unit // "vw"
    }
  });
```

A proper unit i.e. `vw` or `vh` will be picked based on `growDirection`. To encapsulate the logic I moved all dimensions related things to `useUnitHandle()`.

> Note
> Later, we can decide to expose its interface instead of `unit` param, so more complex implementations will be possible.

Existing tests were updated to use proper units. New tests added.


packages/react-resize-handle/src/hooks/useResizeHandle.ts
```diff
@@ -280,12 +292,15 @@ export const useResizeHandle = (params: UseResizeHandleParams) => {
...
         }
+
         attachMouseHandlers(node);
         attachKeyboardHandlers(node);
+
         updateElementsAttrs({
           type: EVENTS.handleRef,
           event: new CustomEvent(EVENTS.handleRef),
-          value: currentValue.current,
+          value: UNMEASURED,
```
layershifter, line 302: There and later, we cannot use stale (old) values if a `ref` was remounted: it's either a new element or options were updated

packages/react-resize-handle/src/hooks/useResizeHandle.ts
```diff
@@ -134,8 +146,8 @@ export const useResizeHandle = (params: UseResizeHandleParams) => {
       });
 
       // Make sure to only apply the value if it's not the initial value!
-      if (currentValue.current !== UNMEASURED) {
-        updateTargetElVariable(`${currentValue.current}px`);
+      if (eventData.value !== UNMEASURED) {
+        updateTargetElVariable(eventData.value);
```
layershifter, line 150: We should use a value from an event, not `currentValue`

packages/react-resize-handle/src/hooks/useResizeHandle.ts
```diff
@@ -64,19 +66,25 @@ export type UseResizeHandleParams = {
    * A function that will be called to get the value that will be set as the aria-valuetext attribute on the resize handle.
    * Use this for localization.
    */
-  getA11ValueText?: (value: number) => string | undefined;
+  getA11ValueText?: (value: number, unit: string) => string | undefined;
```
ling1726, line 69: Curious to know if narration for aria-valuetext works with viewport units

layershifter: image

Works the same way as for pixels:

image

image

But I would not call VoiceOver experience great

packages/react-resize-handle/src/hooks/useUnitHandle.ts
```diff
@@ -0,0 +1,115 @@
...
+        if (windowDimensionsRafId.current) {
+          targetWindow.cancelAnimationFrame(windowDimensionsRafId.current);
+          windowDimensionsRafId.current = null;
+        }
+      };
+    }
+  }, [targetDocument, unit]);
+
+  return React.useMemo(() => {
+    const name: CSSUnitName =
```
ling1726, line 58: instead of if branches in the handle, can we just have separate implementations based on the selected unit?

packages/react-resize-handle/src/hooks/useResizeHandle.ts
```diff
@@ -96,25 +104,29 @@ export const useResizeHandle = (params: UseResizeHandleParams) => {
...
+  const unitHandle = useUnitHandle(growDirection, params.unit ?? 'px');
 
   const updateTargetElVariable = React.useCallback(
-    (value: string) => {
+    (value: number) => {
       const targetEl =
         variableTarget === 'wrapper' ? wrapperRef.current : elementRef.current;
 
-      targetEl?.style.setProperty(variableName, value);
+      targetEl?.style.setProperty(variableName, value + unitHandle.name);
```
ling1726, line 115: instead of string concat here, can we have a `format` function in the handle?

ling1726 (approved):

ling1726 (approved):

PR #375, opened 2025-05-12
fix(resize-handle): do not call `onChange` with stale values

layershifter:


packages/react-resize-handle/playwright.config.ts
```diff
@@ -47,9 +47,5 @@ export default defineConfig({
       name: 'firefox',
       use: { ...devices['Desktop Firefox'] },
     },
-    {
-      name: 'webkit',
-      use: { ...devices['Desktop Safari'] },
-    },
```
layershifter, line 53: Not sure what is actually used as Webkit, but it clearly does not work the same way as Safari on Mac. Disabled it as tests with it were flaky.

packages/react-resize-handle/src/hooks/useResizeHandle.ts
```diff
@@ -154,6 +159,33 @@ export const useResizeHandle = (params: UseResizeHandleParams) => {
         const newValue = clampWithMode(value, minValue, maxValue, relative);
 
         if (newValue !== currentValue.current) {
+          if (!relative) {
```
ling1726, line 162: can we hoist the relative logic and cleanup here to make that code path less relevant like

```ts
if (relative) {
  return relativeHandlingDeprecated(...args);
}

// stuff that works
```

ling1726: also would be nice to have a GH issue for this

ling1726 (approved):

packages/react-resize-handle/src/hooks/useResizeHandle.ts
layershifter: Makes sense, will do 👍

layershifter: Diff became a bit ackward

PR #384, opened 2025-05-13
fix(resize-handle): updates for #375

layershifter:
Updates for #375.

We actually cannot allow CSS variable to exceed `clamp()` and `width` (i.e. `clamp()`) definition could be changed based on, for example, media queries and will cause unexpected jumps. This makes aligned it with the previous implementation except emitting falsy `onChange` events.


ling1726 (approved):

ling1726:
Test?

PR #396, opened 2025-05-21
chore: setup workspaces in monorepo

Hotell:
To follow our other repo setup and recent guidance of nx team, we are enabling `workspaces` for contrib as well.

This adds 2 additional changes:
- dedupe is being resolved against latest versions ( to enable using `@types/react@18` as source of truth without need of resolution config, once integration tests are implemented for lower react versions )
- normalizes transitive dependency usage


package.json
```diff
@@ -60,14 +60,15 @@
     "@storybook/preview-api": "7.6.20",
     "@storybook/react-webpack5": "7.6.20",
     "@storybook/test-runner": "0.14.0",
-    "@storybook/testing-library": "^0.2.2",
```
Hotell, line 63: unused - added RTL/DOM old dupe

packages/nx-plugin/package.json
```diff
@@ -6,15 +6,15 @@
   "generators": "./generators.json",
   "executors": "./executors.json",
   "dependencies": {
-    "@nx/devkit": "^20.2.2",
+    "@nx/devkit": "^20",
```
Hotell, line 9: simplified dep resolution to mitigate churn when migrating to new nx versions in repo

package.json
```diff
@@ -12,7 +12,7 @@
     "type-check": "nx affected --target=type-check",
     "preinstall": "node ./tools/scripts/preinstall.mjs",
     "postinstall": "patch-package",
-    "dedupe": "npx yarn-deduplicate --strategy fewer"
+    "dedupe": "npx yarn-deduplicate --strategy highest"
```
Hotell, line 15: will dedupe to highest possible semver range -> thus all packages in our repo are bumped as part of this PR

dmytrokirpa (approved):

Issue #350, opened 2025-04-02
Generator script add erroneous entries to nx.json and package.json


brandonthomas:
See history of https://github.com/microsoft/fluentui-contrib/pull/349/files for details.

Both nx.json and package.json had erroneous entries in them from the generator. New contributors or even contributors who aren't often here might not realize these are not intended. We should ensure the generation scripts do not falsely lead people down the wrong path.

The following was added to nx.json
``` json
"@nx/js:swc": {
    "cache": true,
    "dependsOn": ["^build"],
    "inputs": ["production", "^production"]
}
```

This line was added into package.json in the root folder:
``` json
"typescript-eslint": "8.25.0",
```

PR #403, opened 2025-05-30
Token analyzer mcp scaffold

brandonthomas:


nx.json
```diff
@@ -38,6 +38,11 @@
       "options": {
         "packageRoot": "dist/packages/{projectName}"
       }
+    },
```
brandonthomas, line 41: I'm not sure if this is expected or not. This entire PR is auto-generated with the library generator.

Hotell: yeah its not please revert https://github.com/microsoft/fluentui-contrib/issues/350

package.json
```diff
@@ -82,6 +82,7 @@
     "globals": "15.15.0",
     "jest": "29.7.0",
     "jest-environment-jsdom": "29.7.0",
+    "jest-environment-node": "^29.7.0",
```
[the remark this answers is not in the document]
brandonthomas: Yup, there's no UI so no need for JSDOM test env.

Issue #402, opened 2025-05-30
KeytipManager API support in v9


meenalaggarwal:
We have a existing feature in v8 fluent to trigger the keytip mode through the API exposed in Keytip Manager.
Similar functionality is required in V9 Fluent Keytip as well

mainframev:
Resolved in #406

PR #406, opened 2025-06-02
feat(react-keytips): add useKeytipsManager hook

### LINKED ISSUE #404: Keytips don't work with React.SrictMode
Keytips don't enter keytip mode when wrapped with `React.StrictMode`.

### Steps to reproduce:
- Open the https://stackblitz.com/edit/etyw6pjj-ckuuq8p9?file=src%2Findex.tsx
- Try to enter keytip mode by pressing `control`+`option` (on MacOS, the combination might be different on Windows)

https://github.com/user-attachments/assets/bda85bce-3879-43e5-8017-485ec17b41ea

### Actual result:
Keytips don't enter keytip mode

### Expected result:

Keytips should enter keytip mode

**Note:** This issue might occur only in dev mode, as React performs rendering twice. https://react.dev/reference/react/StrictMode#enabling-strict-mode-for-entire-app

mainframev:
BREAKING: useEventServices and `EVENTS` constant are no longer part of public component API 

What's changed: 

- `useKeytipsManager` hook. Users sometimes need manually dispatch keytips events.
-  Fixed `useKeytips`. The hook did not have subscribtion on `ENTER_KEYTIP_MODE` event, so manual dispatching did not trigger showing / hiding keytips.
-  Updated docs and added a few additional test cases.
-  Added `uniqueId` proper to `useKeytipRef`

I've cloned stackblitz example repo with StrictMode and tested out this implementation with [yalc](https://www.npmjs.com/package/yalc), after these changes it works fine. Will double check after we will merge this in stackblitz. 

Implements #402 and fixes #404

### DIFF
--- a/packages/react-keytips/stories/Default.stories.tsx
+++ b/packages/react-keytips/stories/Default.stories.tsx
@@ -100,7 +100,6 @@ const MenuButtonComponent = () => {
   const firstMenuItemRef = useKeytipRef<HTMLDivElement>({
     keySequences: ['1b', 'e'],
     content: 'E',
-    uniqueId: 'bur',
     onExecute,
   });
 
@@ -137,7 +136,6 @@ export const DefaultStory = () => {
 
   const normalButton = useKeytipRef({
     keySequences: ['1c'],
-    uniqueId: 'meow',
     content: '1C',
     onExecute,
   });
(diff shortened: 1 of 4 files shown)


dmytrokirpa (commented):
I'm not sure why we chose to export these hooks and events from the package. While this method works, I'm not a fan of the pattern since it feels like an internal detail that's leaked out. I think a better design could be:



```ts
// 1. hook that abstracts the implementation inside
import { useKeytipManager } from '@fluentui-contrib/react-keytips';

function Component() {
  const { enterKeytipMode, exitKeytipMode } = useKeytipManager()
}
 
// 2. no hooks at all, can be used in both class and functional components, or even outside of them
import { enterKeytipMode, exitKeytipMode } from '@fluentui-contrib/react-keytips';
```

mainframev:
> I'm not sure why we chose to export the `useEventService` and `Events` from the package. While this method works, I'm not a fan of the pattern since it feels like an internal detail that's leaked out. I think a better design could be:
> 
> ```ts
> // 1. hook that abstracts the implementation inside
> import { useKeytipManager } from '@fluentui-contrib/react-keytips';
> 
> function Component() {
>   const { enterKeytipMode, exitKeytipMode } = useKeytipManager()
> }
>  
> // 2. no hooks at all, can be used in both class and functional components, or even outside of them
> import { enterKeytipMode, exitKeytipMode } from '@fluentui-contrib/react-keytips';
> ```

I kinda agree with that, the reason is that it still will be an abstraction over useEventService, so I thought we could export only this, as it might not be used often. I can create this hook.

packages/react-keytips/src/index.ts
```diff
@@ -27,4 +27,5 @@ export { sequencesToID } from './utilities';
 export type { KeytipsProps, KeytipsSlots, KeytipsState } from './Keytips';
 
 export { useKeytipRef } from './hooks/useKeytipRef';
-export { useEventService } from './hooks/useEventService';
```
dmytrokirpa, line 30: Isn't this a breaking change?

mainframev: It is, look at the change file, I marked it as breaking. Component is not stable yet, so it's minor change :)

dmytrokirpa: What about `Events` map/object, shouldn't we remove it's export as well?

PR #411, opened 2025-06-04
Added SplitButton teams component

fogelek:
This PR adds the SplitButton component to the Teams component library:
- Simplified usage (No need to wrap split button with MenuTrigger)
- Built-in tooltip
- Extended functionality to allow separate tooltips between primary button and menu button


packages/teams-components/stories/SplitButton/Default.stories.tsx
```diff
@@ -0,0 +1,240 @@
...
+    background: 'red',
+  },
+});
+
+export const Default = () => {
+  const styles = useStyles();
+  return (
+    <div className={styles.sampleContainer}>
+      <Menu positioning="below-end">
+        <SplitButton>Split</SplitButton>
```
ling1726, line 36: Is it accessible to create a split button without a title/aria-label for the menu button part?

packages/teams-components/src/components/SplitButton/SplitButton.test.tsx
```diff
@@ -0,0 +1,30 @@
...
+import { CalendarRegular } from '@fluentui/react-icons';
+
+describe('SplitButton', () => {
+  it('should render', () => {
+    render(<SplitButton>Test</SplitButton>);
+  });
+
+  it('should throw error if menuTitle is provided without main title', () => {
+    expect(() =>
+      render(<SplitButton menuTitle="Menu">Test</SplitButton>)
```
ling1726, line 13: I don't understand this requirement, why would menu title make the title required?

packages/teams-components/src/components/SplitButton/renderSplitButton.tsx
```diff
@@ -0,0 +1,54 @@
...
+  if (titleProps.title === undefined) {
+    return (
+      <state.root>
+        {state.primaryActionButton && <state.primaryActionButton />}
+        {state.menuButton && <state.menuButton />}
+      </state.root>
+    );
+  }
+
+  return titleProps.menuTitle !== undefined ? (
```
ling1726, line 38: nit: let's convert all these into if blocks for readability

packages/teams-components/src/components/SplitButton/SplitButton.test.tsx
fogelek: If title is provided without menu title, tooltip will be applied to the full SplitButton
If both are given, there will be separate tooltip for action button and menu button - so I think it makes no sense to be able to provide only menu title

packages/teams-components/src/components/ToggleButton/renderTooltip.tsx
```diff
@@ -0,0 +1,14 @@
+import * as React from 'react';
+import { Tooltip, TooltipProps } from '@fluentui/react-components';
+import { StrictSlot } from '../../strictSlot';
+
+export const renderTooltip = (
```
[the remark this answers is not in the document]
fogelek: By mistake. Fixed

PR #429, opened 2025-07-03
fix(stylelint-plugin): allow more pseudo classes

layershifter:
This pull request addresses a bug in the `@fluentui-contrib/stylelint-plugin` package by expanding the list of allowed pseudo-classes in the combinator depth rule. The changes ensure better compatibility with modern CSS practices.

### Bug Fix:

* [`packages/stylelint-plugin/src/rules/combinator-depth/combinator-depth.ts`](diffhunk://#diff-dbd92a56311c51dd812fbc72d3f44ded6527efd1081ba25461aa0f0ada775117R7-R8): Introduced a constant `ALLOWED_PSEUDO_CLASSES` to centralize the list of allowed pseudo-classes (`:hover`, `:active`, `:focus`, `:focus-within`) and updated the logic to use this constant when validating selectors. [[1]](diffhunk://#diff-dbd92a56311c51dd812fbc72d3f44ded6527efd1081ba25461aa0f0ada775117R7-R8) [[2]](diffhunk://#diff-dbd92a56311c51dd812fbc72d3f44ded6527efd1081ba25461aa0f0ada775117L58-R60)

### DIFF
--- a/change/@fluentui-contrib-stylelint-plugin-07a48668-e1bc-4b9b-905e-994574b8380a.json
+++ b/change/@fluentui-contrib-stylelint-plugin-07a48668-e1bc-4b9b-905e-994574b8380a.json
@@ -1,0 +1,7 @@
+{
+  "type": "patch",
+  "comment": "fix: allow more pseudo classes",
+  "packageName": "@fluentui-contrib/stylelint-plugin",
+  "email": "olfedias@microsoft.com",
+  "dependentChangeType": "patch"
+}


packages/stylelint-plugin/src/rules/combinator-depth/combinator-depth.ts
```diff
@@ -4,6 +4,8 @@ import { extractAllSelectors } from '../../postcss-utils';
 import { createRule } from '../../create-rule';
 import { normalizeRuleName } from '../../normalize-rule-name';
 
+const ALLOWED_PSEUDO_CLASSES = [':hover', ':active', ':focus', ':focus-within'];
```
ling1726, line 7: could this be updated to a rule config?

PR #431, opened 2025-07-06
fix(react-draggable-dialog): drag ux and position issue

zhongsp:
This PR fixes:

- Bad UX while dragging the `react-draggable-dialog` dialog, since no realtime position feedback was provided.
- A related positioning issue in existing code.

Details:

The prior implementation uses `transform: CSS.Translate.toString(transform),` to reflect the latest dragging position. But this CSS style is ignored, since the `react-dialog` component adds its own `transform` CSS property. So we have to use `top, left` CSS properties instead.

To the repo owners, feel free to edit the PR or create new based on my implementation.


marcosmoura:
The fix will land with the https://github.com/microsoft/fluentui-contrib/pull/493

PR #474, opened 2025-08-11
docs(react-virtualizer): Update virtualizer stories to use the same storybook Meta format as contrib repo

Mitch-At-Work:
Our virtualizer stories currently aren't displayed on the docsite, assuming this is due to storybook type (works on local storybook).

Updating to match the other story setups.

### DIFF
--- a/apps/docsite/src/Welcome.mdx
+++ b/apps/docsite/src/Welcome.mdx
@@ -30,6 +30,8 @@ Explore the various packages available to extend your Fluent UI capabilities:
 
 - [**react-tree-grid**](/docs/packages-react-tree-grid--docs): TreeGrid component for hierarchical data.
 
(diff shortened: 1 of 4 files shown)


packages/react-virtualizer/src/components/Virtualizer/renderVirtualizer.tsx
```diff
@@ -1,6 +1,3 @@
-/** @jsxRuntime automatic */
-/** @jsxImportSource @fluentui/react-jsx-runtime */
```
Hotell, line 2: this will break slots. you need to add following

```
/** @jsxRuntime classic */
/** @jsx createElement */
import { createElement } from '@fluentui/react-jsx-runtime';
```

Hotell (approved):

PR #486, opened 2025-08-28
fix(docsite): add missing dependency for extract-storybook-llms target

dmytrokirpa:
Fixes an issue when the `extract-storybook-llms` target is executed before the `build-storybook` https://github.com/microsoft/fluentui-contrib/actions/runs/17131066622/job/48595224133:

image

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,7 +41,7 @@ jobs:
 
       - run: yarn nx affected --target=verify-integrity
       - run: yarn nx affected --target=build
-      - run: yarn nx affected --target=build-storybook,extract-storybook-llms
+      - run: yarn nx affected --target=build-storybook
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -41,7 +41,7 @@ jobs:
 
       - run: yarn nx affected --target=verify-integrity
       - run: yarn nx affected --target=build
-      - run: yarn nx affected --target=build-storybook
+      - run: yarn nx affected --target=build-storybook,extract-storybook-llms
```
dmytrokirpa, line 44: todo: revert before merge, just to make sure that build passes on CI

mainframev: @dmytrokirpa  do you need to have a build-storybook target here, if extract-storybook-llms is dependent on that and it still will be build?

PR #493, opened 2025-09-09
fix(react-draggable-dialog): improve calculation performance after motion upgrade on Dialog

marcosmoura:
Dialog component, that is used under the hood by the DraggableDialog, was upgraded to use Motion Preview APIs, and that was making the component to misbehave.

This PR adds a bunch of performance optimizations as well, to better calculate the position of the DraggableDialog after the Motion upgrade.

- Closes https://github.com/microsoft/fluentui-contrib/issues/424

### DIFF
--- a/packages/react-draggable-dialog/jest.config.ts
+++ b/packages/react-draggable-dialog/jest.config.ts
@@ -1,15 +1,10 @@
 /* eslint-disable */
 import { readFileSync } from 'fs';
-import { dirname, join } from 'path';
-import { fileURLToPath } from 'url';
-
-const __filename = fileURLToPath(import.meta.url);
-const __dirname = dirname(__filename);
 
 // Reading the SWC compilation config and remove the "exclude"
 // for the test files to be compiled by SWC
 const { exclude: _, ...swcJestConfig } = JSON.parse(
-  readFileSync(join(__dirname, '.swcrc'), 'utf-8')
+  readFileSync(`${__dirname}/.swcrc`, 'utf-8')
 );
 
 // disable .swcrc look-up by SWC core because we're passing in swcJestConfig ourselves.
@@ -19,7 +14,7 @@ if (swcJestConfig.swcrc === undefined) {
 }
 
 // Uncomment if using global setup/teardown files being transformed via swc
-// https://nx.dev/packages/jest/documents/overview#global-setup/teardown-with-nx-libraries
+// https://nx.dev/nx-api/jest/documents/overview#global-setupteardown-with-nx-libraries
 // jest needs EsModule Interop to find the default exported setup/teardown functions
 // swcJestConfig.module.noInterop = false;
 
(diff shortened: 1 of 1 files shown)


packages/react-draggable-dialog/src/components/DraggableDialogSurface/useDraggableDialogSurface.ts
```diff
@@ -109,19 +119,14 @@ export const useDraggableDialogSurface = (
...
-        return;
+        return undefined;
       }
 
       const boundaryTopReference = boundaryRect.top + boundaryRect.height / 2;
       const boundaryLeftReference = boundaryRect.left + boundaryRect.width / 2;
       const top = boundaryTopReference - Math.ceil(currentEl.clientHeight / 2);
       const left = boundaryLeftReference - Math.ceil(currentEl.clientWidth / 2);
 
-      setDropPosition({
```
marcosmoura, line 120: This was causing React to set the state wrongly, especially after removing the element from DOM

packages/react-draggable-dialog/jest.config.ts
```diff
@@ -19,11 +24,11 @@ if (swcJestConfig.swcrc === undefined) {
 // swcJestConfig.module.noInterop = false;
 
 export default {
-  displayName: 'button',
+  displayName: 'react-draggable-dialog',
```
marcosmoura, line 27: Not super relevant for this PR, but this is completely wrong. So, I fixed here.

packages/react-draggable-dialog/package.json
```diff
@@ -3,9 +3,9 @@
   "version": "0.1.2",
   "dependencies": {
     "@swc/helpers": "~0.5.11",
-    "@dnd-kit/core": "^6.1.0",
+    "@dnd-kit/core": "^6.3.1",
     "@dnd-kit/utilities": "^3.2.2",
-    "@dnd-kit/modifiers": "^7.0.0"
+    "@dnd-kit/modifiers": "^9.0.0"
```
marcosmoura, line 8: Upgrade for performance

packages/react-draggable-dialog/src/components/DraggableDialog/useDraggableDialog.ts
```diff
@@ -68,36 +68,52 @@ export const useDraggableDialog = (
...
-      if (!onPositionChange || !rect) {
-        return;
-      }
-
-      const position = {
-        x: rect.left,
-        y: rect.top,
-      };
+      cancelOnDragAnimationFrame();
+      setOnDragAnimationFrame(() => {
```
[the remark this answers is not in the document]
dmytrokirpa: 1. why not debouncing/throttling the update?
2. I'm not 100% sure, but if want to use the animation frame, we'll probably need to make sure that state is not updated after the component unmounted, something like this:
```
React.useEffect(() => {
   return () => cancelOnDragAnimationFrame();
}, [cancelOnDragAnimationFrame])
```

Issue #183, opened 2024-05-29
Bug: fix browser global usage in `houdini-utils`


spmonahan:
There are several instances where `houdini-utils` directly accesses browser globals like `document` and, implicitly, `window` that break when run in multi-window applications.

Issues that need to be resolved are labeled with `// TODO: fix global.`

PR #512, opened 2025-09-26
Avoid global document/window usage as it breaks in multi-window apps

marco-ms:
Fix for #183
Usage of global document/window leads to non working UX when used in multi-window environment


marco-ms:
New pull request available here: https://github.com/microsoft/fluentui-contrib/pull/514

PR #520, opened 2025-10-03
fix(houdini-utils): Give wrapper element a different id 

CampbellOwen:
In #514 the creation of the wrapper element for the fallback canvas was given the same id as the canvas element itself. This means that when the id is used to reference the canvas to display the fallback paint animation it is blank instead.

This change gives the wrapper a stable `houdini-fallback-wrapper-<x>` id


spmonahan (approved):

layershifter:
@spmonahan @CampbellOwen thx, I merged the PR. Please trigger a release ;)

PR #524, opened 2025-10-15
test(react-draggable-dialog): add a lot more test cases for better coverage

marcosmoura:
This PR adds a bunch of new tests to improve the coverage of the React Draggable Dialog.

### DIFF
--- a/packages/react-virtualizer/stories/VirtualizerScrollViewDynamic/ComplexDynamicList.stories.tsx
+++ b/packages/react-virtualizer/stories/VirtualizerScrollViewDynamic/ComplexDynamicList.stories.tsx
@@ -438,4 +438,4 @@ export const ComplexDynamicList = (): JSXElement => {
   );
 };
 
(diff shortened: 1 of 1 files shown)


packages/react-draggable-dialog/src/components/DraggableDialog/utils/getParsedDraggableMargin.test.ts
```diff
@@ -0,0 +1,232 @@
...
+
+  describe('when margin is undefined', () => {
+    it('should return default margin', () => {
+      const result = getParsedDraggableMargin(undefined);
+      expect(result).toEqual(defaultMargin);
+    });
+  });
+
+  describe('when margin is a number', () => {
+    it('should return uniform margins for various number types', () => {
```
dmytrokirpa: You could leverage the [it.each](https://jestjs.io/docs/api#testeachtablename-fn-timeout) and have one assertion per tests with a proper scenario title

```suggestion
  it.each(testCases, 'should return margins for number types: %o %o', (actual, expected) => {
```

PR #544, opened 2025-10-31
fix(react-draggable-dialog): improve initial position calculation performance

marcosmoura:
This PR introduces several performance improvements related to the calculation of styles and position.

- Reduces the number of re-renders and expensive calculation.
- Make Draggabler Dialog responsive when drag is not initiated
- Add an example on how to memo the draggable dialog to improve performance and avoid re-renders.


packages/react-draggable-dialog/src/components/DraggableDialog/useDraggableDialog.ts
```diff
@@ -17,6 +18,9 @@ import {
 } from './DraggableDialog.types';
 import { getParsedDraggableMargin } from './utils/getParsedDraggableMargin';
 import { restrictToBoundaryModifier } from './utils/restrictToBoundaryModifier';
+import { DraggableDialogContextValue } from '../../contexts/DraggableDialogContext';
+
+const noopPositionChange = () => ({});
```
mainframev: it's not in diff, but on line 47 I guess we could use this noopPositionChange 

`onPositionChange: onPositionChange ?? (() => ({}))` -> `onPositionChange: onPositionChange ?? noopPositionChange`

packages/react-draggable-dialog/src/components/DraggableDialog/useDraggableDialog.ts
```diff
@@ -70,6 +74,16 @@ export const useDraggableDialog = (
   const [dropPosition, setDropPosition] = React.useState({ x: 0, y: 0 });
   const [hasBeenDragged, setHasBeenDragged] = React.useState(false);
 
+  const onPositionChangeRef = React.useRef(onPositionChange);
+  React.useEffect(() => {
+    onPositionChangeRef.current = onPositionChange;
+  }, [onPositionChange]);
+
+  const lastReportedPositionRef = React.useRef({
+    x: Number.NaN,
```
mainframev: can't we use here `x: 0, y: 0` as initials?

packages/react-draggable-dialog/src/components/DraggableDialog/useDraggableDialog.test.ts
```diff
@@ -1,11 +1,34 @@
...
+
+const setAnimationFrameSpy = jest.fn((callback: () => void) => callback());
+const cancelAnimationFrameSpy = jest.fn();
+
+fluentComponents.useAnimationFrame.mockImplementation(() => [
+  (callback: () => void) => {
+    setAnimationFrameSpy(callback);
+  },
+  cancelAnimationFrameSpy,
+]);
```
layershifter:
```
import { useAnimationFrame }  from "@fluentui/react-components"

const useAnimationFrameMock = jest.mocked(useAnimationFrame).mockImplementation(...);
```

https://jestjs.io/docs/mock-function-api#jestmockedsource-options

PR #527, opened 2025-10-20
fix(react-virtualizer): fix errors when the items order changes

dmytrokirpa:
This PR fixes item order issues in `react-virtualizer`, resolving React key errors and improving stability. It also adds thorough tests for edge cases.

Fixes #526

PR #545, opened 2025-10-31
fix (react-virtualizer): Ensure items re-render if the render function changes

Mitch-At-Work:
We need items to re-render when the render callback changes, this will ensure updates are completed.

Teams will need to handle minimal re-rendering via useCallback and appropriate dependencies for optimization.


Mitch-At-Work:
Closing this for now in favor of this change instead: https://github.com/microsoft/fluentui-contrib/pull/527

PR #551, opened 2025-11-15
cleanup resize observer during cleanup

micahgodbolt:
### Changes
- Added `resizeObserver` to `FallbackAnimationState` to make it accessible for cleanup
- Call `disconnect()` on ResizeObserver when animation completes, is stopped, or cleaned up
- Prevents observer from holding stale references to `targetWindow` and `targetEl`

### DIFF
--- a/packages/houdini-utils/src/types.ts
+++ b/packages/houdini-utils/src/types.ts
@@ -218,5 +218,5 @@ export type FallbackAnimationState = {
   mode: 'moz-element' | 'webkit-canvas' | 'to-data-url';
   wrapper: HTMLElement | null;
   running: boolean;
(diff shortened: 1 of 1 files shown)


micahgodbolt:
Another related concern is that resize observers on the root window and targetWindow could cause unexpected re-renders. Not sure if this can be avoided or not. I know the houdini utils default to window if targetWindow isn't valid. Is this something we should guard against? Here's a repro copilot came up with (take with grain of salt)

https://stackblitz.com/edit/vitejs-vite-rsxxbbyy?file=index.html&terminal=dev

mainframev (approved):

PR #562, opened 2025-11-25
Update react-cap-theme structure to follow new guidelines

dvdzkwsk:
This updates a few things:

The folder structure is now more consistent with FluentUI where each component has its own folder. It also introduces prefixes to token names to help us track the origins of tokens since we're pulling these from multiple sources.

- `smtc/v1` Indicates a token from semantic tokens v1 / the [MAI library](https://mai.microsoft.design/storybook/?path=/docs/documentation-semantic-tokens-mai-token-values--docs&globals=theme:phoenix-dark). Everything after the prefix is the actual token name copy+pasted.
  - An example would be smtc/v1/background/card/onPrimary/default/disabled.
- `smtc/v2` Indicates a token from the new semantic tokens v2 effort. Everything after the prefix is the actual token name copy+pasted.
  - An example would be smtc/v2/ctrl/avatar/active-ring-size.
- `cap/` Indicates a new token that we've introduced just for ourselves. Eventually, we expect semantic tokens to introduce a token that fulfills the same purpose.
  - An example would be cap/ctrl/button/corner-radius.

For any `cap/` tokens we're adding, we propose following the conventions set by the MAI library: https://mai.microsoft.design/storybook/?path=/docs/documentation-semantic-tokens-introduction--docs, which follows this convention: `category/group/scale/item/sub-item/property/variant/state`. An example would be `ctrl/input/sm/radio/doc/background/primary/rest`.

Importantly, it's not worth twisting ourselves in knots trying to adhere perfectly with this structure. The idea is to try to introduce some structure so that we're not all doing wildly different things, but we fully expect there to still be inconsistencies and that's okay. We can always clean it up later.


sxsx1xsxs (approved):
Testing one approval from cap-theme group

PR #566, opened 2025-11-26
Enhance Storybook for CAP theme selection

pixel-perfectionist:
- Add CAP theme selector toolbar (paintbrush icon with label) to react-cap-theme Storybook, including Teams/OneDrive/SharePoint variants.
- Normalize and extend root Storybook decorators, wrapping stories in CAPThemeSelectionProvider.
- Introduce shared utils for theme selection/context and example rendering across variants.
image

### DIFF
--- a/packages/react-cap-theme/stories/components/Button.stories.tsx
+++ b/packages/react-cap-theme/stories/components/Button.stories.tsx
@@ -14,7 +14,7 @@ const buttonAppearances = [
   'primary',
   'outline',
   'subtle',
-  'transparent'
+  'transparent',
 ] as const;
 
 type ButtonState = 'default' | 'hover' | 'active' | 'focus' | 'disabled';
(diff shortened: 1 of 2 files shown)


sxsx1xsxs (commented):
so a couple of high level feedback:

1. Love the toggle that contains all theme! however, currently it only works for Button. Badge and Card for example will not change the look when the toggle is applied. We need to fix it.
2. button size should not be a global toggle, because it doesn't make sense for the other components

pixel-perfectionist:
> so a couple of high level feedback:
> 
> 1. Love the toggle that contains all theme! however, currently it only works for Button. Badge and Card for example will not change the look when the toggle is applied. We need to fix it.
> 2. button size should not be a global toggle, because it doesn't make sense for the other components

1. I left the other stories as-is to avoid touching code I’m not responsible for and potentially disrupting ongoing work. My changes show the proposed structure, and others can migrate their stories once we align on it.

2. removed it

dvdzkwsk (changes requested):
Need to fix storybook version+change to devDependency and address a few other minor comments.

packages/react-cap-theme/stories/components/Button.stories.tsx
```diff
@@ -9,70 +9,172 @@ import {
...
+                      }}
+                    >
+                      <div style={{ fontWeight: 600, textTransform: 'capitalize', width: '100px', textAlign: 'right', margin: '0 24px 0 0' }}>
+                        {appearance}
+                      </div>
+                      <Button
+                        id={stateId(appearance, 'default')}
+                        icon={<CalendarMonth />}
+                        {...commonButtonProps}
+                        appearance={appearance}
```
mainframev: could you also add `@ts-expect-error` suppression here? RIT tests are failing on CI because there is no `tint` appearance variant in official Button API 

image

PR #585, opened 2025-12-10
[CAP-theme] Add Dialog component

sxsx1xsxs:
This PR adds Dialog component according to https://www.figma.com/design/40dstXfCx3jpvlP4prbSBQ/Semantic-Tokens-mapping?node-id=6104-27312&m=dev

It adds two type of stories
1. table view
2. control view

It also set up a structure where other components could follow to have the two types of stories and add add extra note column to the storybook table.

With Control
Screenshot 2025-12-10 at 3 16 06 PM

Table view
Screenshot 2025-12-10 at 3 15 52 PM

Also updated the other story names so it's clear the story type
Screenshot 2025-12-10 at 3 39 40 PM

### DIFF
--- a/packages/react-cap-theme/stories/index.stories.tsx
+++ b/packages/react-cap-theme/stories/index.stories.tsx
@@ -4,12 +4,12 @@ import { Meta } from '@storybook/react';
 export { CAPBadgeStory as Badge } from './components/Badge.stories';
 export { CAPButtonWithCtrlsStory as ButtonWithControls } from './components/Button.stories';
 export { CAPCardStory as Card } from './components/Card.stories';
+export { CAPDialogStory as Dialog } from './components/Dialog/WithTable';
+export { CAPDialogWithCtrlsStory as DialogWithControls } from './components/Dialog/WithControls.stories';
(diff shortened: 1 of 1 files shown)


packages/react-cap-theme/src/components/Dialog/DialogSurface.styles.ts
```diff
@@ -0,0 +1,29 @@
...
+  type DialogSurfaceState,
+  useDialogSurfaceStyles_unstable,
+} from '@fluentui/react-components';
+
+import { CAP_TOKENS } from '../../theme/CAPTheme';
+
+export const useDialogSurfaceStyles = makeStyles({
+  root: {
+    borderRadius: CAP_TOKENS['cap/Dialog/Corner'],
+    border: `${CAP_TOKENS['cap/Dialog/strokeWidth']} solid ${CAP_TOKENS['cap/Dialog/strokeColor']}`,
```
pixel-perfectionist: @sxsx1xsxs could you update it please to use `shorthands` util, to be consisten to Fluent:
` ...shorthands.borderColor(borderPressed)`

PR #626, opened 2026-04-04
feat: switch to fluentui storybook addons

dmytrokirpa:
This pull request updates Storybook configuration and documentation for several packages to use the official Fluent UI Storybook addons and improves the clarity and maintainability of package documentation. It removes custom Fluent UI Storybook components in favor of the new addons, updates dependencies, and replaces dynamically imported README descriptions with concise, static summaries.


packages/react-tree-grid/stories/index.stories.tsx
```diff
@@ -16,7 +16,10 @@ const meta = {
   parameters: {
     docs: {
       description: {
-        component: [description, a11y].join('\n'),
+        component: [
+          'This package provides a set of styleless components to implement the `treegrid` navigation pattern.',
+          a11y,
```
Hotell, line 21: we should either inline all md docs or keep as is. while the babel gives us warning during compilation these provide no harm for the output. thoughts ?

PR #631, opened 2026-04-09
feat(react-cap-theme): update published components

sxsx1xsxs:


packages/react-cap-theme/src/components/react-accordion/index.ts
GianoglioEnrico, line 1: We don't need to re-export all the Accordion components from @fluentui/react-..
We can only re-export:
```
export { useAccordionHeaderStyles } from "./components/AccordionHeader/useAccordionHeaderStyles.styles";

export { useAccordionPanelStyles } from "./components/AccordionPanel/useAccordionPanelStyles.styles";
```

packages/react-cap-theme/src/components/react-avatar/index.ts
GianoglioEnrico, line 1: Similar to Accordion, we don't need to re-export all the Avatar components.
We can re-export use{Compoent}.styles.ts only.

packages/react-cap-theme/src/components/react-avatar/components/AvatarGroupItem/AvatarGroupItem.types.ts
GianoglioEnrico, line 1: This file is not needed, can be removed.

packages/react-cap-theme/src/components/react-avatar/components/AvatarGroupItem/useAvatarGroupItem.ts
GianoglioEnrico, line 1: This file is not needed, can be removed.

packages/react-cap-theme/src/components/react-avatar/components/AvatarGroupPopover/AvatarGroupPopover.types.ts
GianoglioEnrico, line 1: This file is not needed, can be removed.

packages/react-cap-theme/src/components/react-badge/index.ts
GianoglioEnrico, line 1: We can re-export use{Compoent}.styles.ts only.

packages/react-cap-theme/src/components/react-button/components/Button/useButtonStyles.styles.ts
GianoglioEnrico, line 198: The button base styles hardcode `borderRadius: tokens.borderRadius2XLarge`, but the shape prop (rounded | square | circular) from ButtonState is never read or applied. This means all buttons render as pill-shaped (rounded) regardless of the shape value.
We should add to useRootStyles somenthing like:

```
const useRootStyles = makeStyles({
......
rounded: {
		/* The borderRadius rounded styles are handled in the size variations */
	},
	square: { borderRadius: tokens.borderRadiusNone },
	circular: { borderRadius: tokens.borderRadiusCircular },
....
```


and then:

```
const {
		appearance,
		disabled,
		disabledFocusable,
		iconOnly,
		iconPosition,
		shape,
		size,
	} = state;

state.root.className = mergeClasses(
		....
		rootStyles[shape],
....
```

packages/react-cap-theme/src/index.ts
GianoglioEnrico: You can remove:
```
  useAccordionStyles_unstable: (state) => {
    return useAccordionStyles(state as AccordionState);
  },
  useAccordionItemStyles_unstable: (state) => {
    return useAccordionItemStyles(state as AccordionItemState);
  },
```

GianoglioEnrico: You can remove:
` useAvatarGroupStyles_unstable: (state) => {
    return useAvatarGroupStyles(state as AvatarGroupState);
  },`
We have not CAP styles for this.

GianoglioEnrico (commented):
@sxsx1xsxs Great work on this! Alongside the inline comments I left, a couple of things to address:

Index exports cleanup — Each component's index.ts is currently re-exporting everything from @fluentui/react-components. We should remove those re-exports and only keep the CAP style overrides (e.g., useButtonStyles, useAccordionHeaderStyles). Consumers already get the base components from Fluent directly.

Unused files — Since we're focused on restyling only for now, could you double-check if there are any non-style files that were carried over but aren't actually needed? We want to keep the surface area minimal.

Thanks so much for putting this together🙂

packages/react-cap-theme/src/index.ts
GianoglioEnrico, line 1: @sxsx1xsxs  forgot to mention this in my previous comments — this file should pretty much only contain the imports for component styles and component states(i.e:BadgeState) , along with the export of CAP_STYLE_HOOKS. The rest of the exports (icons, tokens, component files, etc.) aren't needed here and can be safely removed. Thanks!

GianoglioEnrico (approved):

packages/react-cap-theme/src/components/react-button/components/Button/useButtonStyles.styles.ts
```diff
@@ -378,10 +390,13 @@ export const useButtonStyles = (state: ButtonState): ButtonState => {
   state.root.className = mergeClasses(
     state.root.className,
     buttonClassNames.root,
-    rootBaseClassName,
+    rootBaseStyles.base,
+    rootBaseFocusIndicatorStyles.base,
     rootStyles.root,
     rootStyles[appearance],
+    rootStyles[shape],
```
GianoglioEnrico: ` rootStyles[shape]` must come after `rootStyles[size]` otherwise `small` size overrides borderRadius for other shapes.

GianoglioEnrico (approved):

PR #632, opened 2026-04-13
fix(react-cap-theme): remove unused tokens

dvdzkwsk:
For most tokens, we should import from `@fluentui/tokens` directly. Only for _new_ tokens should we import from `tokens.ts`. I've named these `capTokens` for clarity. In the process, I also removed all tokens that weren't referenced anywhere in code.

We don't have any downstream consumers of these tokens right now so backwards compatibility is not a concern.


GianoglioEnrico (approved):
I understand the reason why we're defining the CAP tokens this way. Just to make sure we're aligned and we keep track of what's already in Fluent vs what's not — these "CAP" tokens:
```
borderRadius2XLarge
borderRadius3XLarge
borderRadius4XLarge
colorNeutralStroke4
colorNeutralStroke4Hover
colorNeutralStroke4Pressed
colorNeutralStroke4Selected
```
are [defined and exported in Fluent UI](https://github.com/microsoft/fluentui/pull/35574) starting from version 9.72.10. The Fluent UI version used in the @fluentui-contrib repo is currently 9.70.0, so these tokens aren't available here yet. Teams is already on version 9.73.1, so they are available there.

PR #633, opened 2026-04-14
chore(deps): upgrade @fluentui/react-components v9.70.0 -> v9.73.7

dvdzkwsk:
It seems like this repo prefers to maintain a single version of fluent, so I upgraded it at the root rather than adding a `devDependency` to `packages/react-cap-theme`, which wants access to new tokens in @fluentui/react-components.

### DIFF
--- a/change/@fluentui-contrib-react-cap-theme-e737166c-ae30-49c1-ad87-333cc012402c.json
+++ b/change/@fluentui-contrib-react-cap-theme-e737166c-ae30-49c1-ad87-333cc012402c.json
@@ -1,0 +1,7 @@
+{
+  "type": "none",
+  "comment": "fix react-cap-theme types",
+  "packageName": "@fluentui-contrib/react-cap-theme",
+  "email": "dzukowski@microsoft.com",
+  "dependentChangeType": "none"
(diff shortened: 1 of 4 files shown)


Hotell:
@dvdzkwsk yup we follow single version policy if possible, seem you forgot to commit some changes. ty

dvdzkwsk:
> @dvdzkwsk yup we follow single version policy if possible, seem you forgot to commit some changes. ty

@Hotell it should be good now :)

miceclavea (approved):

sxsx1xsxs (commented):
there is no package.json update but only lock file update?

sxsx1xsxs (approved):

dvdzkwsk:
> there is no package.json update but only lock file update?

Yes. It isn't necessary to update the package.json in this case. Obviously we could, but it isn't necessary.

dvdzkwsk:
@mainframev, mind taking a look? The checks are passing.

dvdzkwsk:
@mainframev, @Hotell, could I get a stamp on this if it looks good to you? Thanks!

PR #634, opened 2026-04-14
feat(react-cap-theme): add react-input component

dvdzkwsk:


packages/react-cap-theme/src/components/react-input/components/Input/useInputStyles.styles.ts
```diff
@@ -0,0 +1,280 @@
...
+    flexWrap: 'nowrap',
+    alignItems: 'center',
+    verticalAlign: 'middle',
+    minHeight: '36px',
+    padding: `${tokens.spacingVerticalNone} ${tokens.spacingHorizontalMNudge}`,
+    color: tokens.colorNeutralForeground1,
+    backgroundColor: tokens.colorNeutralBackground1,
+    border: `${tokens.strokeWidthThin} solid ${tokens.colorNeutralStrokeAccessible}`,
+    borderRadius: capTokens.borderRadius2XLarge,
+    ...typographyStyles.body1,
```
GianoglioEnrico, line 24: In SharePoint, the full Input component was re-exported with all styling applied standalone, rather than calling Fluent's base styles. Since our styles are layered on top of Fluent's base styles instead of replacing them, the base `::after` pseudo-element (the animated focus underline) still comes through:

Image

The quickest fix for now would probably be to add `'::after': { content: 'unset' }` to the root styles.

GianoglioEnrico (approved):

PR #635, opened 2026-04-14
feat(react-cap-theme): add react-label component

dvdzkwsk:


packages/react-cap-theme/src/components/react-label/components/Label/useLabelStyles.styles.ts
```diff
@@ -0,0 +1,100 @@
...
+};
+
+const useStyles = makeStyles({
+  root: {
+    alignItems: 'center',
+    display: 'inline-flex',
+  },
+  large: { fontWeight: tokens.fontWeightRegular },
+  required: { color: tokens.colorStatusDangerForeground3 },
+  semiBold: { fontWeight: tokens.fontWeightSemibold },
```
GianoglioEnrico, line 24: It looks like the Label semibold weight `fontWeight: tokens.fontWeightSemibold` is already applied by default in Fluent UI ([useLabelStyles.styles.ts](https://github.com/microsoft/fluentui/blob/8ad0bbc45c71a850ff7f601d53cb48ebbf0a0903/packages/react-components/react-label/library/src/components/Label/useLabelStyles.styles.ts#L51)), so this override seems redundant.

GianoglioEnrico (approved):

packages/react-cap-theme/src/components/react-label/components/Label/useLabelStyles.styles.ts
dvdzkwsk: Good call out. I'm going to leave this for a future PR that deduplicates local styles w/ what's already in fluent.

PR #640, opened 2026-04-22
feat(react-cap-theme): add TEAMS_STYLE_HOOKS with button size overrides

sxsx1xsxs:
Introduces a `teams/` folder inside `react-cap-theme` that layers Teams-specific overrides on top of `CAP_STYLE_HOOKS`. Currently covers the button family (Button, MenuButton, SplitButton, ToggleButton), restoring V9-compatible heights (Small=24px, Medium=32px) via vertical padding overrides to preserve zoom and reflow behaviour.

`TEAMS_STYLE_HOOKS` is exported from the existing package entry point alongside `CAP_STYLE_HOOKS` 

Figma file (everything noted in this frame is implemented except for 4. Consideration for Dark & High Contrast theme components, which needs further discussion): https://www.figma.com/design/YaBNE9cxfSOBLgWO5qONDA/Teams-Visual-Refresh----Phase1--Full-Spec?node-id=1925-55900&m=dev

https://dev.azure.com/microsoftdesign/ModeDS/_workitems/edit/36971


packages/react-cap-theme/src/teams/components/react-button/components/MenuButton/useMenuButtonStyles.styles.ts
```diff
@@ -0,0 +1,29 @@
...
+  useRootSizeStyles,
+} from '../Button/useButtonStyles.styles';
+
+export const useMenuButtonStyles = (
+  state: MenuButtonState
+): MenuButtonState => {
+  const rootSizeStyles = useRootSizeStyles();
+  const rootIconOnlySizeStyles = useRootIconOnlySizeStyles();
+
+  capUseMenuButtonStyles(state);
```
dvdzkwsk: Hooks should always start with the "use" prefix.

```suggestion
  useCapMenuButtonStyles(state);
```

packages/react-cap-theme/src/teams/components/react-button/components/MenuButton/useMenuButtonStyles.styles.ts
```diff
@@ -0,0 +1,29 @@
+import {
+  useMenuButtonStyles as capUseMenuButtonStyles,
+  type MenuButtonState,
+} from '@fluentui-contrib/react-cap-theme/react-button';
```
dvdzkwsk: It seems strange to me that here we are importing from "@fluentui-contrib/react-cap-theme/react-button", then below we are importing from a relative path "../Button/useButtonStyles.styles". We should stick to just one, which is probably relative paths.

packages/react-cap-theme/src/teams/components/react-button/components/MenuButton/useMenuButtonStyles.styles.ts
```diff
@@ -0,0 +1,29 @@
...
+  state: MenuButtonState
+): MenuButtonState => {
+  const rootSizeStyles = useRootSizeStyles();
+  const rootIconOnlySizeStyles = useRootIconOnlySizeStyles();
+
+  capUseMenuButtonStyles(state);
+
+  state.root.className = mergeClasses(
+    state.root.className,
+    state.iconOnly
```
dvdzkwsk: It isn't clear to me why we need this `iconOnly` condition in useMenuButtonStyles, useSplitButtonStyles, etc. This should be handled already by useButtonStyles, right?

sxsx1xsxs: The `iconOnly` condition is needed in MenuButton/ToggleButton. Teams' `useMenuButtonStyles` calls CAP's hook (which internally calls CAP's `useButtonStyles`, not Teams'). So Teams' `useButtonStyles` iconOnly handling never runs for MenuButton

sxsx1xsxs: from Claude 

```
  When a MenuButton renders with TEAMS_STYLE_HOOKS, Fluent calls only
  useMenuButtonStyles_unstable — it does not separately call
  useButtonStyles_unstable. So the flow is:

  TEAMS_STYLE_HOOKS.useMenuButtonStyles_unstable
    → Teams useMenuButtonStyles
      → useCapMenuButtonStyles(state)          ← CAP's hook
          → CAP's useButtonStyles(state)       ← CAP's, not Teams'

  Teams' useButtonStyles is never in this chain. So the iconOnly size override
  logic from Teams' useButtonStyles never runs for MenuButton.

  The condition is also doing real work — the two style maps are different:

  - rootSizeStyles — only overrides paddingTop/paddingBottom (keeps horizontal
  padding from CAP)
  - rootIconOnlySizeStyles — overrides all four sides of padding +
  minWidth/maxWidth to produce a square button

  Without the iconOnly check, an icon-only MenuButton would get the text-button    
  override (vertical padding only), leaving it with CAP's minWidth: 28px and
  asymmetric padding — non-square and wrong size.  
```

dvdzkwsk:
```
  When a MenuButton renders with TEAMS_STYLE_HOOKS, Fluent calls only
  useMenuButtonStyles_unstable — it does not separately call
  useButtonStyles_unstable. So the flow is:

  TEAMS_STYLE_HOOKS.useMenuButtonStyles_unstable
    → Teams useMenuButtonStyles
      → useCapMenuButtonStyles(state)          ← CAP's hook
          → CAP's useButtonStyles(state)       ← CAP's, not Teams'
```

I understand that this is what's happening, it's just that this chain seems confusing to me. It's weird to me that we have a Teams-specific `useButtonStyles`, but that function is never actually called as part of this chain. Instead, we are piecing together some utility functions. That means if somebody changes `useButtonStyles` in `/teams`, it won't actually have any effect on this component. Whereas changing `useButtonStyles` in the base CAP theme _would_ have an effect. Does that make sense?

sxsx1xsxs: ya that makes sense

sxsx1xsxs: it is very very confusing

sxsx1xsxs: refactored to address this feedback! thx for the feedback

dvdzkwsk (approved):

PR #642, opened 2026-04-22
Update react-cap-theme version

sxsx1xsxs:


sxsx1xsxs:
Should not be manual change

PR #649, opened 2026-05-07
feat(react-cap-theme): add `react-toolbar` component 

GianoglioEnrico:


packages/react-cap-theme/src/components/react-toolbar/components/Toolbar/useToolbarStyles.styles.ts
```diff
@@ -0,0 +1,35 @@
+import { makeStyles, mergeClasses, shorthands } from '@griffel/react';
+import { tokens } from '@fluentui/tokens';
+import { getSlotClassNameProp_unstable } from '@fluentui/react-utilities';
+import type { ToolbarState } from './Toolbar.types';
+
+const useStyles = makeStyles({
+  root: {
+    ...shorthands.border('1px', 'solid', tokens.colorTransparentStroke),
```
miceclavea: could we use a token for the 1px - tokens.strokeWidthThin

miceclavea (approved):

PR #650, opened 2026-05-07
feat(react-cap-theme): `react-menu` components refactoring

GianoglioEnrico:


packages/react-cap-theme/src/components/react-menu/components/MenuItemLink/useMenuItemLinkStyles.styles.ts
```diff
@@ -0,0 +1,72 @@
...
+ * @alpha
+ */
+export const useMenuItemLinkStyles = (
+  state: MenuItemLinkState
+): MenuItemLinkState => {
+  const styles = useStyles();
+
+  state.root.className = mergeClasses(
+    menuItemLinkClassNames.root,
+    state.root.className,
```
miceclavea, line 30: this merge class has the original state not in the first line, is that intentional?

GianoglioEnrico: Good call, let me remove it since it's not needed.

miceclavea (approved):

PR #651, opened 2026-05-07
React-cap-theme: Removed props being omitted from base Fluent

miceclavea:
- **Button**: `Pick` of 4 base props → `Omit` of only `appearance`, exposing all base props like `shape`, `icon`
- **ToggleButton**: `Pick` of `defaultChecked`/`checked` → `Omit` of `appearance`, exposing `isAccessible` etc.
- **MenuButton**: `Pick` of 4 ButtonProps → full `ButtonProps`
- **Tooltip**: Removed `withArrow` from the `Omit` list so consumers can pass it
- **CarouselButton**: `Pick` on slots/props/state → full Fluent types with `Omit<..., 'appearance'>`
- **CarouselAutoplayButton**: Same — `Pick` → full types with `Omit<..., 'appearance'>`
- **CarouselNavContainer**: `Pick<..., 'layout'>` → `Omit` of slot props (includes all non-slot props)

**No changes**: SplitButton (structural slot Omits), Link (already correct), CarouselNavContainer slot Omit (tooltip type swap), State-level `Required<Pick>` patterns.

### DIFF
--- a/packages/react-cap-theme/src/components/react-button/components/MenuButton/MenuButton.types.ts
+++ b/packages/react-cap-theme/src/components/react-button/components/MenuButton/MenuButton.types.ts
@@ -4,7 +4,11 @@ import type { ButtonProps, ButtonState, ButtonSlots } from '../../Button';
 
 export type { MenuButtonSlots } from '@fluentui/react-button';
 
-export type MenuButtonProps = ComponentProps<MenuButtonSlots> & ButtonProps;
+export type MenuButtonProps = ComponentProps<MenuButtonSlots> &
+  Pick<
+    ButtonProps,
+    'appearance' | 'disabled' | 'disabledFocusable' | 'shape' | 'size'
+  >;
 
 export type MenuButtonState = ComponentState<MenuButtonSlots> &
   Omit<ButtonState, keyof ButtonSlots | 'components' | 'iconPosition'>;
--- a/packages/react-cap-theme/src/components/react-button/components/MenuButton/MenuButton.types.ts
+++ b/packages/react-cap-theme/src/components/react-button/components/MenuButton/MenuButton.types.ts
@@ -4,8 +4,7 @@ import type { ButtonProps, ButtonState, ButtonSlots } from '../../Button';
 
 export type { MenuButtonSlots } from '@fluentui/react-button';
 
-export type MenuButtonProps = ComponentProps<MenuButtonSlots> &
-  Pick<ButtonProps, 'appearance' | 'disabled' | 'disabledFocusable' | 'size'>;
+export type MenuButtonProps = ComponentProps<MenuButtonSlots> & ButtonProps;
 
 export type MenuButtonState = ComponentState<MenuButtonSlots> &
   Omit<ButtonState, keyof ButtonSlots | 'components' | 'iconPosition'>;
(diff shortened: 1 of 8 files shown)


packages/react-cap-theme/src/components/react-button/components/MenuButton/MenuButton.types.ts
```diff
@@ -4,8 +4,7 @@ import type { ButtonProps, ButtonState, ButtonSlots } from '../../Button';
 
 export type { MenuButtonSlots } from '@fluentui/react-button';
 
-export type MenuButtonProps = ComponentProps<MenuButtonSlots> &
-  Pick<ButtonProps, 'appearance' | 'disabled' | 'disabledFocusable' | 'size'>;
+export type MenuButtonProps = ComponentProps<MenuButtonSlots> & ButtonProps;
```
GianoglioEnrico: I think here we should have: 
```
export type MenuButtonProps = ComponentProps<MenuButtonSlots> &
  Pick<
    ButtonProps,
    'appearance' | 'disabled' | 'disabledFocusable' | 'shape' | 'size'
  >; 
```
rather than `...MenuButtonSlots & ButtonProps` directly. Fluent's MenuButtonProps explicitly excludes iconPosition and using & ButtonProps would re-introduce it as a phantom prop that compiles but does nothing at runtime.

packages/react-cap-theme/src/components/react-carousel/components/CarouselButton/CarouselButton.types.ts
```diff
@@ -7,16 +7,15 @@ import type { ComponentProps, ComponentState } from '@fluentui/react-utilities';
 import type { ButtonSlots } from '@fluentui/react-button';
 import type { ButtonProps, ButtonState } from '../../../react-button';
 
-export type CarouselButtonSlots = ButtonSlots &
-  Pick<FluentCarouselButtonSlots, 'root'>;
+export type CarouselButtonSlots = ButtonSlots & FluentCarouselButtonSlots;
```
GianoglioEnrico: nit: `FluentCarouselButtonSlots` already extends ButtonSlots, so the & ButtonSlots here is a no-op — this can simplify to:
`export type CarouselButtonSlots = FluentCarouselButtonSlots;`

packages/react-cap-theme/src/components/react-tooltip/components/Tooltip/Tooltip.types.ts
```diff
@@ -10,7 +10,7 @@ import type { ComponentProps, TriggerProps } from '@fluentui/react-utilities';
 export type TooltipProps = ComponentProps<TooltipSlots> &
```
GianoglioEnrico: This can be simplified to just: `export type TooltipProps = BaseTooltipProps & {
  appearance?: 'normal' | 'inverted' | 'brand';
};`

GianoglioEnrico (approved):
Overall the PR looks good to me, so I've approved it. I left a few minor inline comments — the main one worth addressing is the iconPosition prop on MenuButton since that can also create inconsistencies with the tool you're building.

packages/react-cap-theme/src/components/react-button/components/MenuButton/MenuButton.types.ts
miceclavea: That makes sense, good catch! I applied your suggestion.

packages/react-cap-theme/src/components/react-carousel/components/CarouselButton/CarouselButton.types.ts
miceclavea: Done, thanks

packages/react-cap-theme/src/components/react-tooltip/components/Tooltip/Tooltip.types.ts
miceclavea: Done, good catch.

miceclavea:
@microsoft-github-policy-service agree company="Microsoft"

PR #652, opened 2026-05-07
feat(react-cap-theme): add `react-card` components

GianoglioEnrico:


miceclavea (approved):
@GianoglioEnrico - approving, but noting the inconsistencies in  className placement within mergeClasses. If it is not an issue, we can disregard.

PR #660, opened 2026-06-22
Fix wrapped row navigation in react-tree-grid

bsunderhus:
## Summary
Fix TreeGrid level navigation when virtualized rows are wrapped in additional container elements.

## What changed
- update `useTreeGridLevelNavigation` to traverse rows lazily within the current treegrid instead of relying on direct row siblings
- preserve wrapped-row navigation behavior in the level-navigation unit tests
- update the virtualized navigation fixtures to exercise wrapped `TreeGridRow` containers
- add a Beachball patch changeset for `@fluentui-contrib/react-tree-grid`

## Validation
- `yarn nx run react-tree-grid:test -- --testPathPattern=useTreeGridLevelNavigation.test`
- `yarn nx run react-tree-grid:component-test --skipInstall`

### DIFF
--- a/packages/react-tree-grid/src/hooks/useTreeGridLevelNavigation.ts
+++ b/packages/react-tree-grid/src/hooks/useTreeGridLevelNavigation.ts
@@ -31,32 +31,53 @@ const findAdjacentRowAtLevel = (
     return { type: 'none' };
   }
 
-  let sibling: Element | null =
-    direction === ArrowDown
-      ? row.nextElementSibling
-      : row.previousElementSibling;
+  const treeGrid = row.closest<HTMLElement>('[role="treegrid"]');
+  if (!treeGrid) {
+    return { type: 'none' };
+  }
 
-  while (sibling) {
-    if (isHTMLElement(sibling) && sibling.getAttribute('role') === 'row') {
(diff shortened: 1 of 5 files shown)


packages/react-tree-grid/src/hooks/useTreeGridLevelNavigation.ts
```diff
@@ -31,32 +31,53 @@ const findAdjacentRowAtLevel = (
...
-        : sibling.previousElementSibling;
+      direction === ArrowDown ? rowWalker.nextNode() : rowWalker.previousNode()
+  ) {
+    if (!isHTMLElement(sibling)) {
+      continue;
+    }
+
+    const siblingLevel = Number(sibling.getAttribute('aria-level'));
+
+    if (Number.isNaN(siblingLevel)) {
```
mshoho, line 70: Worth confirming: the walker now scans the whole treegrid rather than just direct siblings. If any `role="row"` without an `aria-level` can appear between data rows (e.g. a column-header row), it would hit this `NaN` branch and return `none`, disabling level navigation in that direction. Fine for the current fixtures since every row has a level — just flagging the assumption.

mshoho (approved):

PR #662, opened 2026-06-24
feat(react-cap-theme): introduce style hook to specifically opt into border radius updates

dvdzkwsk:
Adds a new export to `@fluentui-contrib/react-cap-theme` for consumers that want CAP's rounded corners but otherwise stock Fluent v9 styling (no color, spacing, or typography changes). This is currently entirely a completely isolated set of styles, but if this ends up being promising we'd remove these border styles from the core CAP_STYLE_HOOKS styles.

```tsx
import { FluentProvider, webLightTheme } from '@fluentui/react-components';
import {
  mergeStyleHooks,
  CAP_STYLE_HOOKS_ROUNDED_CORNERS,
} from '@fluentui-contrib/react-cap-theme';

const styleHooks = mergeStyleHooks([
  CAP_STYLE_HOOKS_ROUNDED_CORNERS,
  // ...other CAP facets, or your own custom style hooks
]);

export const App = () => (
  <FluentProvider theme={webLightTheme} customStyleHooks_unstable={styleHooks}>
    {/* the rest of your app */}
  </FluentProvider>
);
```


change/@fluentui-contrib-react-cap-theme-9ae06b97-135d-45f6-b610-30ca8ac5e4dc.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "minor",
+  "comment": "Introduce CAP_STYLE_HOOKS_BORDERS_ONLY",
+  "packageName": "@fluentui-contrib/react-cap-theme",
+  "email": "6439050+davezuko@users.noreply.github.com",
```
pixel-perfectionist, line 5: first, I see this. Just wonder: is it because I would like to wait a spam?

pixel-perfectionist (approved):

change/@fluentui-contrib-react-cap-theme-9ae06b97-135d-45f6-b610-30ca8ac5e4dc.json
dvdzkwsk: it's to hide my real email address :). Enable **Keep my email addresses private** in https://github.com/settings/emails.

sxsx1xsxs:
I am a little worried that it's all separated, the radius styles are not filtered from components styles but in a completely separated file. Alternative is we pass in through the argument of facets all the way into the object used in makeStyles in each component styles files to filter based on facets.

dvdzkwsk:
> I am a little worried that it's all separated, the radius styles are not filtered from components styles but in a completely separated file. Alternative is we pass in through the argument of facets all the way into the object used in makeStyles in each component styles files to filter based on facets.

I agree with you about the future ideal state. For now, I intentionally want to avoid refactoring the existing code to remove the duplication until we know this is a pattern we want to continue. It's more of a standalone experiment for right now.

